PR #5, opened 2021-01-15
add missing const to src/adt.h

svenpeter42:
adt_getprop_copy is defined as const in src/adt.c but not in src/adt.h

### DIFF
--- a/src/adt.h
+++ b/src/adt.h
@@ -75,8 +75,8 @@ const void *adt_getprop(const void *adt, int nodeoffset, const char *name,
                         int *lenp);
 const void *adt_getprop_length(const void *adt, int nodeoffset,
                                const char *name, int len);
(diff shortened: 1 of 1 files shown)


Ferdi265:
That `const` on the return type actually doesn't do anything (at least in C), and on higher warning levels, compilers actually warn for that (`-Wignored-qualifiers` to be exact, it's part of `-Wextra`).

It doesn't do any harm, but it is a little bit inconsistent IMO, so I guess the "clean" thing to do is to remove the `const` in both the header and C file. It can be kept like this as well, just wanted to point out that technicality.

svenpeter42:
Ah, true. I just fixed the inconsistency to make it compile with clang which complained about a redeclaration with a different type. Feel free to remove the const - no hard feelings from my side ;)

PR #7, opened 2021-01-15
adt: fix ignored qualifiers in adt_getprop_copy function prototype

Ferdi265:
This commit removes the unnecessary `const` on `adt_getprop_copy()` in both the header and the C file, since `const int` as a return type does not do anything different than just `int`, and the added `const` might be confusing to some.

See also the discussion in #5 "add missing const to src/adt.h".


marcan:
This happened because the prototype was copied and pasted from a `const void *` and I changed the type without removing the const (but only one one end) :)

PR #12, opened 2021-01-16
Signedness: Replaces int to uint32_t and size_t for unsigned values

MrSparc:
Signedness: 
- `adt.c`: Replaces `int` usage to `uint32_t` for property size (`lenp`) and `size_t` for property name length (`len`).  
This change differs from the API in libfdt but is consistent with the data types of the property and their usage.
- adt.h: Removes `adt_getprop_length` declaration without implementation

Increases compiler warnings to expose some common issues to review:
- Makefile: `Wsign-compare` / `Wunused-parameters` 
-  UNUSED macro to suppress this warning in function parameters that are in ongoing work, debug or subject to API review.

Signed-off-by: Ariel Machado <ariel@fe.up.pt>

### DIFF
--- a/src/adt.h
+++ b/src/adt.h
@@ -4,6 +4,8 @@
 #define XDT_H
 
 #include <stdint.h>
+#include <stddef.h>
+#include "types.h"
 
 #define ADT_ERR_NOTFOUND 1
 #define ADT_ERR_BADOFFSET 4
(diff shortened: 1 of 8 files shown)


src/adt.h
```diff
@@ -37,6 +38,7 @@ int adt_check_header(const void *adt);
 
 static inline int adt_first_property_offset(const void *adt, int offset)
 {
+    (void)(adt);
```
marcan, line 41: This should be UNUSED

marcan (changes requested):

src/adt.h
MrSparc: Sure, but that implies to include utils.h where the macro is defined just for this line. I was expecting this argument be used in near future or removed otherwise.  But I can change this

marcan: This argument will never be used. It is present for API uniformity, but the way the format works the first property offset is at a constant offset, so no data needs to be read.

MrSparc: Got it. I've been seeing the use of the typedefs of `types.h` in the project (u8, u32, etc.). If the UNUSED macro is moved from utils.h to types.h, by including it in adt.h we can also use these typedefs in the adt code and standardize the use of these typedefs. Do you agree to do it and use this same PR?

marcan:
Yeah, I think that macro makes more sense in types.h to be honest :)

marcan (approved):

PR #15, opened 2021-01-23
MMU support

svenpeter42:
This PR adds basic MMU support to m1n1.

Still missing:

- Tests on real hardware. This has only been verified inside qemu so far
- Proxy calls to disable the MMU before e.g. jumping to Linux


MrSparc:
My suggestion is to use the defined types (aliases u8, u32, u64, etc.) in `types.h` so that their use be consistent throughout the code.

PR #16, opened 2021-01-28
Add a few details to build on debian buster to README

amworsley:
just add a few things I had work out to get things compiling


baryluk:
Nah. This is not needed or specific to debian. Just clone with `--recursive`. I updated the instruction on the wiki.

marcan:
I've updated the README, hope it makes more sense now :)

PR #17, opened 2021-01-31
utils.h: Fix 'asm-operand-widths' warnings with clang

MrSparc:
Clang raises 'asm-operand-widths' warnings in inline assembly code when the size of an operand is < 64 bits and the operand width is unspecified.
Warnings are raised in `_msr(reg, val)` macro, i.e. the datatype of the operand may vary. 
In the macro we cast the operand to 64 bits, forcing the use of the right 64-bit register size.

Signed-off-by: Ariel Machado <ariel@fe.up.pt>

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -215,8 +215,12 @@ static inline u8 mask8(u64 addr, u8 clear, u8 set)
     })
 #define mrs(reg) _mrs(reg)
 
-#define _msr(reg, val) ({ __asm__ volatile("msr\t" #reg ", %0" : : "r"(val)); })
-#define msr(reg, val)  _msr(reg, val)
+#define _msr(reg, val)                                                                             \
+    ({                                                                                             \
+        u64 __val = (u64)val;                                                                      \
+        __asm__ volatile("msr\t" #reg ", %0" : : "r"(__val));                                      \
+    })
+#define msr(reg, val) _msr(reg, val)
 
 #define sysop(op) __asm__ volatile(op ::: "memory")
 


src/utils.h
```diff
@@ -215,8 +215,12 @@ static inline u8 mask8(u64 addr, u8 clear, u8 set)
     })
 #define mrs(reg) _mrs(reg)
 
-#define _msr(reg, val) ({ __asm__ volatile("msr\t" #reg ", %0" : : "r"(val)); })
-#define msr(reg, val)  _msr(reg, val)
+#define _msr(reg, val)                                                                             \
+    ({                                                                                             \
+        u64 __val = (u64)val;                                                                      \
```
baryluk, line 220: Double underscore prefix is reserved to be only used by C compiler and library peoples. ;) C standard permits arbitrary breakage of stuff with double underscore prefixes, so don't use it.

MrSparc: I know all identifiers begin with either two underscores or an underscore followed by a capital letter are reserved names.
But this is a general practice when used by macros inside a block scope  '{' and '}' . This approach is widely used by Linux kernel, some examples :
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/include/asm/sysreg.h#n1076

baryluk: Fair enough.

marcan:
Thanks! Worth noting that we *are* the C library here, so whatever the C spec says about reserved identifiers is less relevant :)

PR #18, opened 2021-01-31
Fix broken hexdumps in proxy.py

modwizcode:
Proxy.py has a few typos that appear if you try to use debug mode and dump a region of memory. This should fix those.


modwizcode:
Forgot the sign off so I fixed that.

PR #21, opened 2021-02-05
Linux loader xz and bootargs support

jannau:
Decompression time is negligible compared to the loading over the 1.5mpbs serial interface.


marcan:
Nice work, thanks! I'd been meaning to clean up the script :)

PR #23, opened 2021-02-07
tooling: Add container setup for build environment

bastilian:
This allows compiling m1n1 on any platform that supports containers / Docker and doesn't require using a separate Linux build machine.

With Docker(-Compose) installed building m1n1 can be done by running: 

```
$ docker-compose build
$ docker-compose run m1n1 make
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM debian:buster-slim
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y build-essential bash git locales gcc-aarch64-linux-gnu libc6-dev-arm64-cross device-tree-compiler imagemagick \
+    && rm -rf /var/lib/apt/lists/* \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+ENV LANG en_US.utf8
+
+WORKDIR /m1n1
+COPY . .
+
+CMD ["/bin/bash"]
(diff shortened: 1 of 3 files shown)


noahwilliamsson:
@bastilian Nice work.  Some nits:

1. to avoid build breakage with future Debian releases, consider using `FROM debian:buster` over `FROM debian:latest`
2. `FROM debian:buster-slim` results in a smaller initial download[1] (the build pulls in the same amount of packages with both `buster` and `buster-slim`)
3. if you use `WORKDIR /m1n1` (dir created automatically) (and `COPY . .`) instead of `RUN mkdir /m1n1` (and `COPY . /m1n1`), users won't need to `cd /m1n1` later
4. if you use `CMD ["/bin/bash"]` over `ENTRYPOINT "/bin/bash"`, it will be possible to `docker run --rm -v the_image $PWD:/m1n1 make` (with `ENTRYPOINT` that command would drop you into a shell)

[1] sample image sizes
```
% docker images
REPOSITORY         TAG          IMAGE ID       CREATED              SIZE
debian             10-slim      91748b3bad73   3 weeks ago          63.4MB
debian             10           db7272368d6f   3 weeks ago          108MB
debian             latest       db7272368d6f   3 weeks ago          108MB
```

bastilian:
@noahwilliamsson I'll update the Dockerfile incorporating the suggested changes. I just quickly put it together to get it built.

I was thinking maybe, since the plan is to use Arch linux, to use it for this as well already, but that was just a side-thought.

svenpeter42:
Very nice!

Small nit: there are two Signed-off-bys in your commit.

bastilian:
/me 🤦 rarely used "signed-off" in the past.

Thank you for the hint @svenpeter42!

marcan:
I'm not much of a Docker person, but I'll trust you on this one that it works and people should find it useful :)

PR #25, opened 2021-02-12
Add a simple frame buffer console

svenpeter42:
- [x] cleanup code some more
- [x] add font license and sign-off the commit
- [x] integrate makefont.sh into Makefile
- [x] figure out how to deal with the logo


src/fb.c
```diff
@@ -40,24 +43,147 @@ static inline uint32_t rgbx_to_rgb30(u32 c)
...
+
+static u32 font_get_pixel(u8 c, u32 x, u32 y)
+{
+    c -= 0x20;
+    u8 *ptr = &_binary_font_b_start[c * 512 + (y * 8 + x) * 4];
+    u8 r = *ptr++;
+    u8 g = *ptr++;
+    u8 b = *ptr++;
+
+    return (b << 2) | (g << 12) | (r << 22);
```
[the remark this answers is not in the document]
modwizcode: You need to use a memcpy to do that safely but otherwise I think that's correct.

src/fb.c
```diff
@@ -1,36 +1,39 @@
...
 #include "fb.h"
+#include "string.h"
 #include "utils.h"
 #include "xnuboot.h"
 
-u32 *fb;
-int fb_s, fb_w, fb_h, fb_d;
-int scale = 1;
+#define CONSOLE_MAX_ROWS 60
+#define CONSOLE_MAX_COLS 80
```
modwizcode, line 11: These need to be calculated based on the width and height of the screen and the font size, otherwise your writes will go out of bounds (and things won't display on the screen properly).

If it's a Mac Mini with a 800x600 monitor connected you're going to run out of bounds.

src/fb.h
```diff
@@ -5,13 +5,23 @@
...
 
-extern u32 *fb;
-extern int fb_s, fb_w, fb_h;
+typedef struct {
+    u32 *ptr;
+    int s;
+    int d;
+    int w;
+    int h;
+} fb_t;
```
modwizcode, line 14: This needs to use real variable names and have a description as to what it's doing. Especially since s here is stride/4 which isn't actually the stride value anymore since it's not the number of bytes per line. It would probably be better to just keep track of the original bootargs.video information directly.

src/fb.c
```diff
@@ -40,24 +43,147 @@ static inline uint32_t rgbx_to_rgb30(u32 c)
...
+void fb_init(void)
+{
+    fb.ptr = (void *)cur_boot_args.video.base;
+    fb.s = cur_boot_args.video.stride / 4;
+    fb.w = cur_boot_args.video.width;
+    fb.h = cur_boot_args.video.height;
+    fb.d = cur_boot_args.video.depth & 0xff;
+    printf("fb init: %dx%d (%d) [s=%d] @%p\n", fb.w, fb.h, fb.d, fb.s, fb.ptr);
+
+    if (cur_boot_args.video.depth & 0x10000) {
```
modwizcode, line 179: While you're cleaning this up you might as well give this flag a proper name (It's for determining if it's a retina screen).

Makefile
```diff
@@ -93,6 +93,10 @@ build/%.o: build/%.bin
 	@echo "  BIN   $@"
 	@$(OBJCOPY) -I binary -B aarch64 -O elf64-littleaarch64 $< $@
 
+build/font.o: font.b
+	@echo "  BIN   $@"
+	@$(OBJCOPY) -I binary -B aarch64 -O elf64-littleaarch64 $< $@
+
```
modwizcode, line 99: Since this is exactly the same as the existing .bin rule call it font.bin and let it handle that automatically using the .bin rule. You can combine this with moving the font generation in as a makefile rule generating a .bin out of a ttf.

src/fb.c
```diff
@@ -1,36 +1,39 @@
...
-
-    if (cur_boot_args.video.depth & 0x10000) {
-        logo = (void *)_binary_build_bootlogo_256_bin_start;
-        logo_w = logo_h = 256;
-    } else {
-        logo = (void *)_binary_build_bootlogo_128_bin_start;
-        logo_w = logo_h = 128;
-    }
-}
+extern u8 _binary_font_b_start[0];
```
modwizcode, line 36: If you change the Makefile like I suggest you will need to rename this to `_binary_font_bin_start`

modwizcode (changes requested):
I think things are mostly good other than a few style and commenting tweaks here.

Also this isn't SMP safe as you've noticed, and you'll probably need to add some form of locking to make sure access to the framebuffer's console state isn't racey.

src/fb.c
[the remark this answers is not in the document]
maximus64: There are other place in this code that does this. So I assume HW take care of unaligned access. I agree that this way is safer. Maybe fix it for fb_blit() as well so it consistent?

marcan:
> I think things are mostly good other than a few style and commenting tweaks here.
> 
> Also this isn't SMP safe as you've noticed, and you'll probably need to add some form of locking to make sure access to the framebuffer's console state isn't racey.

SMP boot already should *effectively* synchronize with the CPUs booting, so we can probably get away without locking here at this time (the UART code has this problem too), but it would be ideal to add locking globally to printf to solve this - however, there are subtleties there with IRQs, so we probably need to disable IRQs around printf to avoid deadlocks there.

modwizcode:
> > I think things are mostly good other than a few style and commenting tweaks here.
> > 
> > Also this isn't SMP safe as you've noticed, and you'll probably need to add some form of locking to make sure access to the framebuffer's console state isn't racey.
> 
> SMP boot already should *effectively* synchronize with the CPUs booting, so we can probably get away without locking here at this time (the UART code has this problem too), but it would be ideal to add locking globally to printf to solve this - however, there are subtleties there with IRQs, so we probably need to disable IRQs around printf to avoid deadlocks there.

Other people noted that testing on physical hardware was causing SMP startup to fail with the new print code. I was mostly referencing that, although I'm not positive what the actual cause if that was.

marcan:
Something to look out for is that right now we map the FB as cached, though given that this code works I assume the GPU is cache-coherent with the CPUs. But since the other CPUs will have caches disabled, and in fact we never enable the MMU on those, I could see this going wrong if different CPUs try to access the same memory with different caching attributes.

src/fb.c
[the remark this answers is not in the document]
svenpeter42: Why is a memcpy required here?

src/fb.c
```diff
@@ -40,24 +43,147 @@ static inline uint32_t rgbx_to_rgb30(u32 c)
...
 }
 
 void fb_display_logo(void)
 {
     printf("fb: display logo\n");
-    fb_blit((fb_w - logo_w) / 2, (fb_h - logo_h) / 2, logo_w, logo_h, logo, logo_w);
+    fb_blit((fb.w - logo.w) / 2, (fb.h - logo.h) / 2, logo.w, logo.h, logo.ptr, logo.w);
 }
+
+void fb_init(void)
```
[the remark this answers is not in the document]
svenpeter42: Because I eventually need to call `fb_console_blit_all();` after the frame buffer is initialised to show the text that was printed before. I could either add a prototype for `fb_console_blit_all()` at the beginning or just move the init further below. I'll fix this in the header for consistency as well.

src/fb.h
svenpeter42: I'll use better variables.

svenpeter42:
> Something to look out for is that right now we map the FB as cached, though given that this code works I assume the GPU (and display controller) are cache-coherent with the CPUs. But since the other CPUs will have caches disabled, and in fact we never enable the MMU on those, I could see this going wrong if different CPUs try to access the same memory with different caching attributes.

Ack, I think that's what is causing these failures.

> 
> Given that, it might be prudent to disable FB prints entirely from CPUs other than CPU#0; that sounds like it's likely to save us a whole bunch of trouble anyway. We can either just not show those messages on the FB, or buffer them and catch up on CPU#0 on the next print from there.

Good idea, I'll just disable frame buffer printfs from secondary cores entirely.

src/fb.c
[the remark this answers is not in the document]
modwizcode: Yeah the latter. Just for aliasing, since even in C you can't just use a pointer cast legally to do that (a union can work but the details are confusing). The bootargs struct is already done like this, other code seems to forget, but it of course should be optimized into a pure load. We could also compile with "-fno-strict-aliasing" which is what linux does but yeah. I hate this issue because I feel like the cast is so much clearer.

src/fb.h
```diff
@@ -5,13 +5,30 @@
...
+    u32 height; /* height of the framebuffer in pixels */
+} fb_t;
+
+typedef struct {
+    u8 r;
+    u8 g;
+    u8 b;
+} rgb_t;
+
+extern fb_t fb;
```
modwizcode, line 22: Same nitpick as fb_t declaration in fb.c, we don't use this anywhere else so it really doesn't have to be a global like that.

src/fb.c
```diff
@@ -1,63 +1,275 @@
...
+    u8 text[CONSOLE_MAX_ROWS][CONSOLE_MAX_COLS];
+} console = {.initialized = 0,
+             .cursor = {.row = 0, .col = 0},
+             .n_rows = CONSOLE_MAX_ROWS,
+             .n_cols = CONSOLE_MAX_COLS,
+             .row_offset = 0,
+             .text = {{0}}};
 
 extern char _binary_build_bootlogo_128_bin_start[0];
 extern char _binary_build_bootlogo_256_bin_start[0];
```
modwizcode, line 63: While you're working on this you might as well change these to match (using `u8` and `[]` instead of `char` and `[0]`)

src/fb.c
```diff
@@ -1,63 +1,275 @@
...
+            memcpy(console.text[row], console.text[row + row_offset], CONSOLE_MAX_COLS);
+        console.cursor.row = console.n_rows;
+    }
+    fb_check_scroll();
+
+    console.initialized = 1;
+    fb_console_blit_all();
+}
+
+void fb_blit(u32 x, u32 y, u32 w, u32 h, void *data, u32 stride)
```
modwizcode, line 249: The stride here is a pixel stride for the source image right? I think that parameter should be before the `void* data` parameter. 
Also it's not really a `void*` since you only ever use it as a `u8*`, and all the things you pass in are also defined that way so I think it's best to not have it be a void*

modwizcode (changes requested):
Few nitpicks, but overall besides the one change in the `fb_console_init` function to fix the bug with row offsets I think this is pretty good. Works well in the emulator even with different size displays and everything.

I like the way it splits the text at the logo. It's simple and prevents the logo from being overwritten or anything. I was going to suggest changing that a bit, I think that just using a small logo in the upper right corner when using fb console might be nice (maybe display the big logo, delay 0.5 sec then replace with small logo and just use the full screen?) but that's a personal preference thing (I think it makes split hex values a little annoying to read).

src/fb.c
```diff
@@ -1,63 +1,275 @@
...
+    u32 n_cols;
+    u32 row_offset;
+
+    u8 text[CONSOLE_MAX_ROWS][CONSOLE_MAX_COLS];
+} console = {.initialized = 0,
+             .cursor = {.row = 0, .col = 0},
+             .n_rows = CONSOLE_MAX_ROWS,
+             .n_cols = CONSOLE_MAX_COLS,
+             .row_offset = 0,
+             .text = {{0}}};
```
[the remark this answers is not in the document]
svenpeter42: In general I keep public interfaces/structs/etc. in .h files with the expectation that these won't change. Internal structure on the other hand are kept inside .c files. These structure are only used internally and I'd like to keep them there.

src/fb.c
```diff
@@ -1,63 +1,275 @@
...
+        console.cursor.row = 0;
+    }
+
+    /*
+     * this size of our console may have shrunk and we'll need to throw away some early rows
+     * and copy later ones to the start.
+     * we can't use fb_check_scroll() here because we will need to copy rows that are already
+     * larger than console.n_rows.
+     */
+    if (console.cursor.row > console.n_rows) {
```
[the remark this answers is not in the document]
svenpeter42: This works fine the way it is. `fb_check_scroll` is meant to restore the invariant `row < n_rows` when being called with `row <= n_rows`.
It will do `n_row--` before hitting that assert. 

I could make this `>=` and add a `+1/-1` in some of the lines below but I think it's easier to read this way. No strong opinion though.

src/fb.c
svenpeter42: good point, will change these as well.

src/fb.h
svenpeter42: If we make fb private we might as well make fb_blit etc. private because you can't tell the width and height of the fb anymore. this could be cleaned up in a separate change but I'll leave it the way it is right now since there are already a lot of changes at this point.

src/fb.c
svenpeter42: same argument like making fb static: probably a good idea but feels like it's out of scope for this change.

modwizcode: I don't really see how it's out of scope, this function's parameters were changed in this series of commits, that's where the stride parameter was introduced.

svenpeter42: No, the stride parameter has always been there. the only change was `int` -> `u32`.

src/fb.c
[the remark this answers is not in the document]
modwizcode: I was encountering an issue where that assert was getting triggered. You're right that the case I mention shouldn't actually happen and I think maybe it was just a coincidence or something else was off. So this is fine.

src/fb.c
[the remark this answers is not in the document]
modwizcode: That's fine then, I think the biggest confusing thing is that the closing brace for console is not on it's own line, which visually makes that section look wrong to me. But that's just personal preference (and it might be correct bracing I didn't check elsewhere), so I guess it's okay to leave as is.

src/fb.c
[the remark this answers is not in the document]
svenpeter42: Did you maybe by chance run an outdated version? I accidentally removed the `fb_check_scroll` call while rebasing and then later ran into that assert because `row` was `n_rows + 1`.

src/fb.c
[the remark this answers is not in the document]
svenpeter42: Yeah, I'd prefer to have the brackets a little bit different as well but this is what `clang-format` did and I just accepted it 😄

src/fb.h
modwizcode: Fair point. I was mostly going off how nothing else _currently_ uses it and wasn't thinking about the fact that it was meant to be part of the public interface for potential future use. In that case it's probably fine.

src/fb.c
modwizcode: You're right I'm not sure how I missed that. My bad sorry about that.

src/fb.c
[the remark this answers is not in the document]
modwizcode: I bet that's what happened. I was using a very weird merge briefly and I thought I had reset everything clean but I think I ended up with thing in a weird state.

src/fb.c
svenpeter42: No worries, it's probably github's fault. The diff for all commits together looks very confusing while the diff for the individual commit looks fine.

modwizcode:
Github's review system is very strange and seems to break completely when someone force pushes (I would remove my previous change requests or like to mark resolved ones that have been discussed), but overall I think once those last few changes are made this is probably good to merge. Thanks for putting up with my confusion @svenpeter42.

svenpeter42:
thanks for the review and the tests on qemu @modwizcode! those definitely improved the code :)

maximus64:
I tried out on my Macbook Air and SMP boot is working. Font also much bigger than before heh. lgtm

marcan:
The terminal looks good; the font looks hideous. drmr would've never let us get away with that in BootMii :-)

I figured out what went wrong with the converter, so now we're good.

Before:

image

After:

image

marcan:
Hm, the "logo avoidance" is kind of wonky:

image

(This is with Retina mode forced on a non-Retina output).

Looks like you need to check the right side of the character, not the left. But... do we really want this kind of weird "wrap-through" behavior?

Personally, when I was thinking about the console, I was just considering limiting the console to the left half of the screen, never going beyond the left edge of the logo. I can see how this approach lets the logo be where it is while letting use use more width but... it feels kind of odd.

One nice thing about not playing games with the logo is that you can scroll by just doing block memmoves; you need to do it linewise to avoid moving the logo, but at least you don't need to repaint every character.

I've also noticed that, as you would expect, once the MMU shuts down the framebuffer becomes Linux-on-crappy-GPU-fbcon-in-the-2000s-quality slow. This is chiefly due to the scroll; there are only a couple lines being printed at that point. This might or might not be improved by doing row-wise scrolling, but either way, can we add a hack for this? Something like a console function that scrolls without moving the true cursor, opening up, say, 3 lines, so that the MMU code can call that prior to MMU shutdown in anticipation that there will be one or two lines printed after that, so we avoid having to scroll.

marcan:
Also, aren't we cropping off long lines instead of wrapping, the way it's done now?

It feels wrong dealing with the console as a buffer of text lines that then get blitted as-is. Text needs to be wrapped (on a character boundary, not word wrap) as it is "reshaped" into console size. If we do scrolling the way I suggested, we don't need to keep the concept of text lines around; writing text just becomes a matter of keeping track of the cursor position, blitting characters, and moving down on newline/overflow, and scrolling once we reach the bottom edge.

To print early text before console init, at that point, you could just buffer it as plain text (though you might still want to count backwards the console height to avoid wasting time rendering lines that will scroll off, but you still might waste a bit of time with overflow). Then it would be wrapped at the right width when the console is initialized and the backlog rendered.

modwizcode:
These are good points that I assumed were somewhat unimportant in my review because I can’t see how much more output would be required and the issue with wrapping around the logo is honestly triggered very rarely. However you’re right, we can simply blit one character at a time and it fixes every issue. It also lets you just overwrite only the part of the logo that you would encroach onto, which was going to be my suggestion originally before realizing it was hard to implement without a different rendering approach.

Is the reason for the lagginess with the MMU turned off related to that disabling caching/write combining? I’ve always wondered why otherwise fast processors struggle with rendering high-res console text.

marcan:
Yes, the lag is what happens when the framebuffer is uncached, which it is on many platforms (M1 is actually kind of an exception since it is a unified memory coherent system).

marcan:
Oh, one more thing: there should be a proxy command to enable/disable the fbcon. Obviously printing to the FB adds latency, and that might slow down some experiments (e.g. where exception handlers are spamming dumps) or be otherwise undesirable (e.g. if we're playing with the framebuffer anyway).

svenpeter42:
fixed in https://github.com/AsahiLinux/m1n1/pull/38

PR #26, opened 2021-02-13
chainload.py: transparently compress payload to improve loading time

svenpeter42:

### DIFF
--- a/proxyclient/chainload.py
+++ b/proxyclient/chainload.py
@@ -39,7 +39,7 @@ for cmd in obj.cmds:
         end = min(len(payload), cmd.args.fileoff + cmd.args.filesize)
         size = end - cmd.args.fileoff
         print("Loading %d bytes from 0x%x to 0x%x" % (size, cmd.args.fileoff, dest))
-        iface.writemem(dest, payload[cmd.args.fileoff:end], True)
+        u.compressed_writemem(dest, payload[cmd.args.fileoff:end], True)
         if cmd.args.vmsize > size:
             clearsize = cmd.args.vmsize - size
             print("Zeroing %d bytes from 0x%x to 0x%x" % (clearsize, dest + size, dest + size + clearsize))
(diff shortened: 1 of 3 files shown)


proxyclient/chainload.py
```diff
@@ -31,6 +32,23 @@
     # Fall back to proxy heap, which will be at the right place in old versions
     new_base = u.memalign(0x10000, memory_size)
 
+def compressed_writemem(iface, dest, data, progress):
```
marcan, line 35: This sounds generically useful, can we move it to utils and make it not leak memory? :-)

proxyclient/chainload.py
```diff
@@ -31,6 +32,23 @@
...
+    if not len(data):
+        return
+
+    payload = lzma.compress(data)
+    compressed_size = len(payload)
+
+    try:
+        compressed_addr = p.malloc(compressed_size)
```
marcan, line 43: Just always use u.malloc; p.malloc is only useful when the data needs to outlive the execution of the script. I expect all python-side allocations to use the python-side allocator by default (it's faster too).

marcan (changes requested):

proxyclient/chainload.py
svenpeter42: Good idea! fixed.

proxyclient/chainload.py
svenpeter42: I just blindly copied the code from above without understanding what was going on ;)
Fixed as well.

proxyclient/utils.py
```diff
@@ -94,6 +94,21 @@ def inst(self, op):
...
+        payload = lzma.compress(data)
+        compressed_size = len(payload)
+        compressed_addr = self.malloc(compressed_size)
+
+        self.iface.writemem(compressed_addr, payload, progress)
+        decompressed_size = self.proxy.xzdec(compressed_addr, compressed_size, dest, len(data))
+
+        assert decompressed_size == len(data)
+
+        self.free(compressed_addr)
```
marcan, line 110: Maybe try/finally guard this? Remember that malloc/free are Python-side, so e.g. if a proxy command fails and breaks m1n1 free will still work.

svenpeter42: good idea, done.

marcan: Nice, I like it!

marcan (approved):

PR #28, opened 2021-02-14
MMU: correctly map MMIO/PCIe ranges

svenpeter42:
Previously all MMIO was mapped twice with different attributes
which may or may not lead to strange behaviour when the same
physical range is accessed from both mappings.

We now have a better idea which ranges require nGnRE and nGnRnE
and can just do it correctly instead.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -269,17 +269,27 @@ static void mmu_add_mapping(uintptr_t from, uintptr_t to, size_t size, u8 attrib
 static void mmu_add_default_mappings(void)
 {
     /*
-     * create MMIO mapping as both nGnRnE (identity) and nGnRE (starting at
-     * 0xf0_0000_0000)
+     * create MMIO mappings. PCIe has to be mapped as nGnRE while MMIO needs nGnRnE.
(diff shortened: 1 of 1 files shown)


marcan:
The double mapping is *how* I tested what is either of those types. Can we keep two extra aliases, one for either type? Since these memory ranges are device memory, the CPU isn't allowed to speculate into them, so having aliasing should not cause any weird behavior if those aliases are not tickled.

I do think it makes sense to do the 1:1 mapping the "correct" way, as it makes the code better match what e.g. u-boot will do.

svenpeter42:
> The double mapping is _how_ I tested what is either of those types. Can we keep two extra aliases, one for either type? Since these memory ranges are device memory, the CPU isn't allowed to speculate into them, so having aliasing should not cause any weird behavior if those aliases are not tickled.

Good point, fixed.

 
> I do think it makes sense to do the 1:1 mapping the "correct" way, as it makes the code better match what e.g. u-boot will do.

Yes, agreed. That's why I wanted to make the 1:1 mapping "correct".

PR #29, opened 2021-02-15
Performance monitor experiments

jannau:
adds Apple system register definitions to proxyclient and handles PMU FIQ interrupts in exception.c

PMU interrupts are considered spurious unless PMU_EXPERIMENTS is defined. Has anyone a better idea than `#ifdef`s to separate core bootloader functionality from experiments to understand the system?


marcan:
Sorry, we stepped over each other a bit; I had already added the basic PMU handling in my local copy. Just pushed that :)

For additional experiments: I think the broader question here is, how do we want to handle IRQs when interacting with Python? It is useful to know when a FIQ fired, but presumably you don't need to actually do the PMC handling in m1n1; you can just mask the FIQ and then make sure Python finds out and then it can poll for the registers. I think it should be fine if Python just polls for the PMC overflow status bit for this too, right?

There is currently one problem that needs fixing: IRQs can interrupt UART proxy reply writes, and break them. We should be disabling IRQs/FIQs around proxy UART reply handling to fix this. Currently, if Python is polling for something in a tight loop waiting for an IRQ, it is quite likely that the IRQ debug message will break the polling. However, this only matters if FIQs are actually unmasked; that's only really relevant when experimenting with the IRQ/FIQ functionality itself, since once you know how those work we can just look at the flag bits in the specific blocks that assert IRQ/FIQ, without actually taking an exception to the vector.

So unless you have use cases that *need* to happen in FIQ/IRQ context, I see a few paths for this:

* Do nothing in particular, and just declare that Python experiments shouldn't really be using or depending on IRQs/FIQs for the most part, other than for basic recon about how those mechanisms work for any given peripheral
* Build some kind of flag or event queue mechanism, like the current exc_count, so Python can find out when an IRQ or FIQ fired, and ideally also what type (AIC reason or GIC for IRQ, PMC/IPI/timer for FIQ) by simple polling.
* Build a proper asynchronous event, a new message type that gets sent on IRQ or FIQ and can be handled like a true interrupt all the way into Python. This would really be the same as the current IRQ status text stuff, but machine-encoded so that the proxy stuff in the Python side can present a nice API for it, maybe even with proper callbacks to register "IRQ handlers" in Python and everything.

In fact, since m1n1 currently doesn't itself use IRQs/FIQs at all, for the third case this could be accomplished end to end by simply having the IRQ/FIQ handlers mask DAIF.[F] or DAIF.[I] in the saved context, sending a message to the proxyclient, and returning; the IRQ/FIQ would then be left active and you could just have a true IRQ handler in Python deal with it and re-unmask that interrupt. Then we'd have two IRQ handling modes: the current "standalone" one which just prints a message and masks whatever (which is convenient as a safe fallback), and a "blocking" mode that only does pure python handling, sending a message and leaving IRQ or FIQ masked until Python deals with it and re-unmasks it.

modwizcode:
I was just thinking of posting an issue with the idea of adding “hook points” for python to use (or at least to make ASM injection easier) as a potential solution. But working on a PoC quickly got into messy things like symbol lookups. Your suggestion is much simpler and better.

In fact it almost already works like this if you don’t care about FIQs because the IRQ handler print reads the reason which masks the IRQ, and nothing is around to potentially automatically unmask it. I think the changes required to make the rest of it work aren’t too terribly annoying, if you don’t plan to work on this @marcan maybe I’ll take a crack at it tomorrow.

leeehui:
@jannau  does this need any other AIC related configuration?  for arm pmu we need GIC PPI related config for interrupt handle

jannau:
@leeehui , I've have only seen interrupts with setting SYS_APL_PMCR0 [10:8] to FIQ mode (4). I haven't looked into AIC if it's possible to enable PMU interrupts through AIC. PMU is possiblly hardwired to FIQ on later SoC including the M1.

PR #31, opened 2021-02-17
Add writeread functions + proxy ops

svenpeter42:
These functions all perform a store direcly followed by a load.
This is useful to e.g. useful to find busy bits which might
already be cleared a few cycles after a write.

### DIFF
--- a/proxyclient/proxy.py
+++ b/proxyclient/proxy.py
@@ -308,6 +308,10 @@ class M1N1Proxy:
     P_MASK32 = 0x111
     P_MASK16 = 0x112
     P_MASK8 = 0x113
+    P_WRITEREAD64 = 0x114
+    P_WRITEREAD32 = 0x115
+    P_WRITEREAD16 = 0x116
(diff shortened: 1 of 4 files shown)


modwizcode (approved):
LGTM, I like these ops, it's a good idea to have them!

Any thoughts about relying on directly embedding the ops into a single assembly block inside each method? Or do you think assuming inlining is good enough?

svenpeter42:
No strong opinion tbh. 

I did look at the disassembly and they were always inlined but who knows what all compiler out there do. But on the other hand I like to avoid inline assembly unless it's absolutely required.

marcan:
I think the point here is to avoid the Python round-trip, not so much having the read/write instructions be back to back. And besides, with the ridiculous speculation/OoO that these things do, a few instructions in between probably amount to zero cycles difference.

Which does mean I'm not sure how useful the actual C functions are here, but they don't hurt, so this look good to me. If we ever have a reason to provide stronger guarantees here (e.g. disable IRQs around them) we can do so then.

PR #32, opened 2021-02-19
Makefile: allow to override ARCH

svenpeter42:
I need to run `ARCH=aarch64-unknown-linux-gnu- PATH="/Users/speter/asahi/aarch64-unknown-linux-gnu/bin/:$PATH" make $@` due to my toolchain which won't work unless I can override `ARCH` in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-ARCH := aarch64-linux-gnu-
+ARCH ?= aarch64-linux-gnu-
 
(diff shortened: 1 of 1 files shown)


jannau:
Gnu make supports overriding variables from the command line. `make ARCH=aarch64-unknown-linux-gnu-` works as expected for me.

svenpeter42:
Ah, good point. That works for me as well.

PR #38, opened 2021-03-15
framebuffer console v2

svenpeter42:
~draft because this is currently based on the usb branch~ independent of the usb branch now


svenpeter42:
> Don't you need the "-interligne-spacing -1" on this command ?
> Same applies for next line with -2.

That's what I'd expect as well, but for some reason using `interline-spacing` on mac os creates a font where some chars have a different height. The commands from the READMe are exactly was I used to generate the font. 
This is essentially the whole reason for adding the generated fonts: The conversion process seems to have subtle differences depending on the OS and/or the imagemagick version.

jfbortolotti:
Ok, got it. I remember this discussion now.

Le sam. 20 mars 2021 à 10:47, Sven Peter ***@***.***> a
écrit :

PR #41, opened 2021-03-16
tunables: add initial version.

svenpeter42:
There are at least two types in the ADT related to USB,
but there's a decent chance that there are even more
required for other devices:

 * A simple tunable that applies to a whole device node
   and all it's MMIO ranges specified in the "reg" property.
   This one seems to just be mask32.

 * A slightly more complex tunable that applies to a single
   MMIO range specified in the "reg" property. So far I've
   only seen 32 bit masks but the format looks like it should
   also support 8,16 and 64 bit masks.

Open question: How do we expose proxyclient ops for these? Unlike everything else they take a string as an argument and right now there's no way to easily serialize those.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ LIBFDT_OBJECTS := $(patsubst %,libfdt/%, \
 
 OBJECTS := adt.o bootlogo_128.o bootlogo_256.o chickens.o exception.o exception_asm.o fb.o \
 	heapblock.o kboot.o main.o memory.o memory_asm.o payload.o proxy.o smp.o start.o startup.o \
-	string.o uart.o uartproxy.o utils.o utils_asm.o vsprintf.o wdt.o $(MINILZLIB_OBJECTS) \
+	string.o tunables.o uart.o uartproxy.o utils.o utils_asm.o vsprintf.o wdt.o $(MINILZLIB_OBJECTS) \
 	$(TINF_OBJECTS) $(DLMALLOC_OBJECTS) $(LIBFDT_OBJECTS)
 
 DTS := t8103-j274.dts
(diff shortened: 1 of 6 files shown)


marcan:
Looks good; for exposing these when we need them, we have the same problem with `kboot_set_bootargs()`. I expect strings to just be poked into memory by python, so the m1n1 side doesn't need to do anything special. However, that is done ad-hoc in `linux.py` right now. What I would suggest is augmenting the general proxy code so it accepts `str` (gets encoded as utf-8 and null-terminated) and `bytes` arguments generically, and does the util malloc/unmalloc for you. Minor layering violation here, as util is nominally dependent on proxy and not the other way around; I can figure something out here.

If you want to add proxy ops, feel free to just do it with raw pointers for now like all the other ops, and once that's in I can cook something up for the auto string copying.

kettenis:
The implemented functions work for most of the PCIe-related tunables.  However, some of those tunables are properties of the child nodes whereas the "reg" property that contains the base address is in the parent node. One way to address this is to walk back up the tree when attempting to locate the base address.

svenpeter42:
> The implemented functions work for most of the PCIe-related tunables. However, some of those tunables are properties of the child nodes whereas the "reg" property that contains the base address is in the parent node. One way to address this is to walk back up the tree when attempting to locate the base address.

Interesting. I assume you're talking about e.g. the "apcie-config-tunables" in the "pci-bridge1" node?

kettenis:
> Interesting. I assume you're talking about e.g. the "apcie-config-tunables" in the "pci-bridge1" node?

Indeed

Doesn't necessarily need to be addressed at this point, but I'd be interested in what your take on handling those would be.

svenpeter42:
Alright, I've just added a function that takes a base address for the local tunables as discussed on IRC

kettenis:
Cool, that fits my needs very well (and works).

marcan:
Works for me!

marcan (approved):

PR #42, opened 2021-03-19
proxyclient/i2c: experiment with the i2c and ti usb chip

svenpeter42:
unfortunately not enough to bring up usb in host mode...


marcan:
Can you reformat this with spaces? See the .editorconfig for reference.

svenpeter42:
sure, done.

PR #43, opened 2021-04-06
ci: add clang-format check to the CI

svenpeter42:


marcan:
Looks like this is missing `sysinc/*.h`. Can we turn this into a Makefile rule (`make format-check`?) and just call that instead, so it's next to the format command? :)

svenpeter42:
sure!

marcan (approved):

PR #46, opened 2021-04-11
pcie: minimal initialization

kettenis:
Minimal initialization of the PCIe hardware such that the tunable
can be applied such that they don't have to be passed along in
the devicetree.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -8,6 +8,7 @@
 #include "heapblock.h"
 #include "memory.h"
 #include "payload.h"
+#include "pcie.h"
 #include "pmgr.h"
 #include "smp.h"
 #include "string.h"
@@ -82,6 +83,7 @@ void m1n1_main(void)
     print_info();
     wdt_disable();
     pmgr_init();
+    pcie_init();
 
     printf("Initialization complete.\n");
 
--- a/src/pcie.c
+++ b/src/pcie.c
@@ -1,0 +1,228 @@
+/* SPDX-License-Identifier: MIT */
+
+#include "adt.h"
+#include "pcie.h"
+#include "pmgr.h"
+#include "tunables.h"
+#include "utils.h"
+
+/*
+ * The ADT uses 17 register sets:
+ *
+ * 0:  90000000 00000006 10000000 00000000  ECAM
+ * 1:  80000000 00000006 00040000 00000000  RC
+ * 2:  80080000 00000006 00090000 00000000  PHY
+ * 3:  800c0000 00000006 00020000 00000000  PHY IP
+ * 4:  8c000000 00000006 00004000 00000000  AXI
+ * 5:  3d2bc000 00000000 00001000 00000000  fuses
+ * 6:  81000000 00000006 00008000 00000000  port 0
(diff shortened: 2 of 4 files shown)


kettenis:
This is pretty much the bare minimum needed to apply the tunables from the ADT.  More code will be needed to actually bring up the links to the PCIe devices.  Without that only the bridge devices for RC ports will show up.

In order to write this code I studied the code released by Corellium, but no code has been copied. Many initialization steps in the Corellium code seem to be unecessary and/or have no observable effect and those steps have been left out.  We may have to revisit these steps later when we learn more about the hardware. For now the hardware seems to be perfectly usable in u-boot and OpenBSD.

Merging this pull request probably needs a little bit of coordination such that that the corresponding u-boot changes can be committed as well. The current u-boot will hang when initializing PCIe.

src/pcie.c
```diff
@@ -0,0 +1,180 @@
...
+    /*
+     * Initialize PHY.
+     */
+
+    if (tunables_apply_local("/arm-io/apcie", "apcie-phy-tunables", 2))
+        return -1;
+
+    /* CLK0REQ/CLK0ACK */
+    set32(phy_base, 1);
+    if (poll32(phy_base, 4, 4, 50000))
```
svenpeter42, line 103: minor nit: if we know the register and bit names I'd `#define` them at the top

src/main.c
```diff
@@ -60,6 +61,7 @@ void m1n1_main(void)
     print_info();
     wdt_disable();
     pmgr_init();
+    pcie_init();
```
svenpeter42, line 64: and another nit: I think this can fail silently right now. I'd either add `printf` for each possible cause in `pcie.c` or just a single `if (pcie_init() < 0) printf("...");` here

svenpeter42 (commented):
Very nice work! Do you have a u-boot branch pushed somewhere that is already compatible with this code?

kettenis:
> Very nice work! Do you have a u-boot branch pushed somewhere that is already compatible with this code?

I have now:
https://github.com/kettenis/u-boot/tree/apple-m1-m1n1-pcie

src/pcie.c
kettenis: I'm not 100% sure if all the names are correct and I'd have to come up with a meaningful name for the register too...
But yes, these need to be #defines.

src/main.c
kettenis: This depends a bit on where m1n1 is going.  Adding printf calls in pcie.c is probably the easiest way out for now. That can be easily converted to returning apropriate error codes and doing something sensible with those later.

svenpeter42:
nice, just tested it and can confirm that this works 👍

src/pcie.c
kettenis: Added #defines for the register and bit names that we have some clues about.

src/main.c
kettenis: Since the usb code prints error messages itself I made pcie do the same.

svenpeter42:
looks good to me now

marcan:
Sorry for taking this long on this one, looks good to me :)

PR #47, opened 2021-04-14
Add USB DWC3 Serial Support

svenpeter42:


README.md
```diff
@@ -107,3 +107,6 @@ licensed under the [OFL-1.1](3rdparty_licenses/LICENSE.OFL-1.1) license and copy
 
 * Copyright 2010-2019 Adobe (http://www.adobe.com/), with Reserved Font Name 'Source'. All Rights Reserved. Source is a trademark of Adobe in the United States and/or other countries.
 * This Font Software is licensed under the SIL Open Font License, Version 1.1.
+
+m1n1 embeds portions of the 2013 dwc3 usb linux driver, which is [BSD-or-GPLv2 dual-licensed](3rdparty_licenses/LICENSE.BSD-3.dwc3) and copyright
```
marcan, line 111: Can we put a direct link to the original file here, on the kernel.org gitweb? That helps keep the history/copyright trail, beyond the headers. It's more important for this one given the rather varied kernel licensing.

svenpeter42: sure, done.

pipcet:
I've been playing around with this code (patching together things from various of @svenpeter42's branches), and it appears to work most of the time, but I must confess I accidentally did a very silly thing and tried to run it with the USB-C port connected to the power supply. 

How confident are we that sort of thing can't cause hardware damage? It did cause an immediate reboot, so it might be a good idea to try to detect this case if we can.

I'm not sure this applies to the code in this PR, but I sometimes did have to reboot because the USB interface didn't come up; the good news is that I could work around the problem by re-initializing the USB devices after a timeout (the bad news is that some changes are required to make reads non-blocking so timeouts work without interrupts...)

The version I'm running boots a Mach-O payload after a timeout, so it allows both experimentation and everyday use of the device without having to boot into 1TR all the time, which I think I would like to open PRs for as soon as the ACM code has been merged (I don't have the serial cable or a way to access the UART).

svenpeter42:
> I've been playing around with this code (patching together things from various of @svenpeter42's branches), and it appears to work most of the time, but I must confess I accidentally did a very silly thing and tried to run it with the USB-C port connected to the power supply.
> 
> How confident are we that sort of thing can't cause hardware damage? It did cause an immediate reboot, so it might be a good idea to try to detect this case if we can.

disclaimer: I've only been using this with my Mac Mini so far and I generally shouldn't be trusted with hardware ;)

My understanding is that all low-level power configuration is handled by the TI chip which isn't even brought up with this code. Without that the ports should never behave as a power source even if everything else goes wrong.
The power cable and another USB host both will also try to sink power (possibly even only after USB PD negotiation) into the macbook air and I don't see any obvious way this could cause actual hardware damage.

It's much more likely that the code just panicked somewhere and triggered the reboot. If you want to try again you could replace reboot() with an infinite loop and hope that some exception appears on the framebuffer.

> 
> I'm not sure this applies to the code in this PR, but I sometimes did have to reboot because the USB interface didn't come up; the good news is that I could work around the problem by re-initializing the USB devices after a timeout (the bad news is that some changes are required to make reads non-blocking so timeouts work without interrupts...)

Essentially this code has two annoying issues that can't be easily worked around. I suspect you ran into either of these:

* It only works once per usb port. If you disconnect the host machine and connect it again the dwc3 controller will not trigger the disconnected/connected event. If you just reset the port you can bring it up again though. The same problem also happens when I use the Linux dwc3 driver in device mode or the Linux xhci driver.  I believe Mac OS relies on events triggered by the TI chip connected over I2C for this to work. That requires more yak shaving though: ideally a GPIO interrupt driver and a i2c driver to read and handle events generated by the TI chip to then manually reset the port.
* Enumeration is somewhat unstable. If the handle_events function is not called often enough within an small timeframe after the driver is initialized the host will likely time out. Fixing this is rather annoying without introducing real interrupt support which would just call handle_events on each dwc3 interrupt though.

pipcet:
> disclaimer: I've only been using this with my Mac Mini so far and I generally shouldn't be trusted with hardware ;)

I understand this is all at my own risk.

> My understanding is that all low-level power configuration is handled by the TI chip which isn't even brought up with this code. Without that the ports should never behave as a power source even if everything else goes wrong.
> The power cable and another USB host both will also try to sink power (possibly even only after USB PD negotiation) into the macbook air and I don't see any obvious way this could cause actual hardware damage.
> 
> It's much more likely that the code just panicked somewhere and triggered the reboot. If you want to try again you could replace reboot() with an infinite loop and hope that some exception appears on the framebuffer.

I did, and it appears to work now. It's likely this was a bug introduced by my own experiments, sorry for the noise.

> > I'm not sure this applies to the code in this PR, but I sometimes did have to reboot because the USB interface didn't come up; the good news is that I could work around the problem by re-initializing the USB devices after a timeout (the bad news is that some changes are required to make reads non-blocking so timeouts work without interrupts...)
> 
> Essentially this code has two annoying issues that can't be easily worked around. I suspect you ran into either of these:
> 
>     * It only works once per usb port. If you disconnect the host machine and connect it again the dwc3 controller will not trigger the disconnected/connected event.

> If you just reset the port you can bring it up again though.

I think "wait for a few seconds to see whether a connection is made, otherwise reset the port a few times, then boot the payload if no one's connecting" would be a good default mode of operation for m1n1. That would avoid this issue, and also the issue I was seeing which might be different.

>     * Enumeration is somewhat unstable. If the handle_events function is not called often enough within an small timeframe after the driver is initialized the host will likely time out. Fixing this is rather annoying without introducing real interrupt support which would just call handle_events on each dwc3 interrupt though.

My impression is the complexity of interrupt handling is something to be avoided, but I'm not sure where you're going with m1n1, to be honest :-)

svenpeter42:
> I think "wait for a few seconds to see whether a connection is made, otherwise reset the port a few times, then boot the payload if no one's connecting" would be a good default mode of operation for m1n1. That would avoid this issue, and also the issue I was seeing which might be different.

That's a neat hack but if the issue is different I'd actually like to understand the root cause and see if it can be fixed. 
Do you use a Linux host and can look at the dmesg output when this happens? 

> 
> > ```
> > * Enumeration is somewhat unstable. If the handle_events function is not called often enough within an small timeframe after the driver is initialized the host will likely time out. Fixing this is rather annoying without introducing real interrupt support which would just call handle_events on each dwc3 interrupt though.
> > ```
> 
> My impression is the complexity of interrupt handling is something to be avoided, but I'm not sure where you're going with m1n1, to be honest :-)

Interrupts actually aren't all that complex at this point and m1n1 already has like 90% of the required code. The reason this code doesn't use interrupts is because @marcan wants to use it for his mmioproxy and we can't rely on interrupts there because the AIC will presumably be controlled by the Mac OS guest.

If enumeration is still an issue once marcan refactors uartproxy.c though it might make sense to just add interrupt support and fall back to calling handle_event manually once mmioproxy starts.

marcan:
Code looks good to me for merging, let's iterate on it after that. @pipcet I agree with sven that I don't think we can cause any hardware damage by just driving the USB peripheral; the Ace2 comes up in UFP-only mode and without more configuration it should never try to drive power out or anything dangerous like that.

Handling reconnects is an interesting one. I think it should work without talking to the Ace2; AIUI I didn't have any problem re-enumerating (in DFU mode or whatever) after hotplugging only the D+/D- pins (which the Ace2 would just be passing through), so there should be some way to get that status from the dwc3 or the phy. Either way though, this isn't a blocker.

svenpeter42:
fwiw, I didn't see any interesting bits flipped in the dwc3 or xhci MMIO space when using `RegMonitor` after either device or host PHY bringup when reconnecting the usb cable. It's clearly something non-standard since Linux seems to have trouble as well.

I didn't look at the PHY register space though. My PHY init sequence is also.. er... minimal so maybe something is missing there.

PR #49, opened 2021-04-16
proxy: linux.py: allow to load u-boot before Linux

svenpeter42:
right now u-boot can bring up the PCIe bus such that Linux only requires
a very minimal driver. this new --uboot flag for linux.py allows to load
u-boot before the Linux kernel to make use of this.
This requires a small hack to inject the booti command into the default
u-boot boot environment.


kettenis:
This is a clever hack that might break whenever u-boot gets updated.  I don't think this should be advocated as *the* way to boot Linux using u-boot, but I can see this being useful to boot a Linux kernel quickly while doing development.

"If it breaks you get to keep both pieces."

svenpeter42:
Yeah, completely agreed. I'm just using this to quickly iterate kernels through u-boot.

marcan:
As far as I'm concerned everything in proxyclient is unstable / for development / might break at any time, so stuff like this is fine :)

PR #51, opened 2021-04-17
chainload.py: reconnect after reboot (useful for USB)

jfbortolotti:
Try to reconnect on exception: needed for USB link.

Note: not tested on UART.

Signed-off-by: Jean-Francois Bortolotti <jeff@borto.fr>


marcan:
Hi, this should be fixed in fcd103b (this is why I had the wait_boot() function, it was just missing this logic). Please let me know if this works for you :)

PR #52, opened 2021-04-20
Mach-O payload support

pipcet:
This isn't particularly useful by itself, but I have another patch which modifies m1n1 so it tries to run the proxy first, then falls back to loading the payload only after it times out. That, I think, is useful, since it means you can have a bootable M1 that you can still start with a different kernel.

### DIFF
--- a/src/macho.c
+++ b/src/macho.c
@@ -1,0 +1,84 @@
+/* SPDX-License-Identifier: MIT */
+
+#include "macho.h"
+#include "malloc.h"
+#include "memory.h"
+#include "utils.h"
+#include "xnuboot.h"
+
+#include "libfdt/libfdt.h"
+
+extern u64 boot_args_addr;
+extern void mmu_shutdown(void);
+
+static void (*macho_start_pc)(u64) = NULL;
+int (*macho_boot)(void);
+
+int macho_boot_impl(void)
+{
+    mmu_shutdown();
+
+    macho_start_pc(boot_args_addr);
+
+    panic("macho call returned\n");
+}
+
+void *macho_load(void *start, size_t size)
+{
+    UNUSED(size);
+    struct macho_header *header = start;
+    struct macho_command *command = (void *)(header + 1);
+    struct macho_command *last_command = (void *)command + header->cmdsize;
+    u64 pc = 0;
+    u64 vmbase = 0;
+    u64 vmtotalsize = 0;
+    while (command < last_command) {
+        switch (command->type) {
+            case MACHO_COMMAND_UNIX_THREAD:
+                pc = command->u.unix_thread.pc;
+                break;
+            case MACHO_COMMAND_SEGMENT_64: {
+                u64 vmaddr = command->u.segment_64.vmaddr;
+                u64 vmsize = command->u.segment_64.vmsize;
+
+                if (vmbase == 0)
+                    vmbase = vmaddr;
+                if (vmsize + vmbase - vmaddr > vmtotalsize)
+                    vmtotalsize = vmsize + vmaddr - vmbase;
+                break;
+            }
+        }
+        command = (void *)command + command->size;
+    }
+    void *dest = memalign(0x10000, vmtotalsize);
+    memset(dest, 0, vmtotalsize);
+    command = (void *)(header + 1);
+    void *virtpc = NULL;
+    while (command < last_command) {
+        switch (command->type) {
+            case MACHO_COMMAND_SEGMENT_64: {
+                if (vmbase == 0)
+                    vmbase = command->u.segment_64.vmaddr;
+                u64 vmaddr = command->u.segment_64.vmaddr;
+                u64 vmsize = command->u.segment_64.vmsize;
+                u64 fileoff = command->u.segment_64.fileoff;
+                u64 filesize = command->u.segment_64.filesize;
+                u64 pcoff = pc - vmaddr;
+
+                memcpy(dest + vmaddr - vmbase, start + fileoff, filesize);
+                if (pcoff < vmsize) {
+
+                    if (pcoff < filesize) {
+                        virtpc = dest + vmaddr - vmbase + pcoff;
+                    }
+                }
+            }
+        }
+        command = (void *)command + command->size;
+    }
+
+    macho_start_pc = virtpc;
+    macho_boot = macho_boot_impl;
+
+    return NULL;
+}
--- a/src/macho.h
+++ b/src/macho.h
@@ -1,0 +1,42 @@
+/* SPDX-License-Identifier: MIT */
+
+#ifndef MACHO_H
+#define MACHO_H
+
+#include "types.h"
+
+#define MACHO_COMMAND_UNIX_THREAD 0x05
+#define MACHO_COMMAND_SEGMENT_64  0x19
+
+struct macho_header {
+    u32 irrelevant[5];
+    u32 cmdsize;
+    u32 irrelevant2[2];
+};
+
+struct macho_command {
+    u32 type;
+    u32 size;
+    union {
+        struct {
+            u32 thread_type;
+            u32 length;
+            u64 regs[32];
+            u64 pc;
+            u64 regs2[1];
+        } unix_thread;
+        struct {
+            char segname[16];
+            u64 vmaddr;
+            u64 vmsize;
+            u64 fileoff;
+            u64 filesize;
+            u64 unused2[2];
+        } segment_64;
+    } u;
+};
+
+void *macho_load(void *start, size_t size);
+extern int (*macho_boot)(void);
+
+#endif
--- a/src/payload.c
+++ b/src/payload.c
@@ -4,6 +4,7 @@
 #include "assert.h"
 #include "heapblock.h"
 #include "kboot.h"
+#include "macho.h"
 #include "smp.h"
 #include "utils.h"
 
@@ -19,6 +20,7 @@ const u8 xz_magic[] = {0xfd, '7', 'z', 'X', 'Z', 0x00};
 const u8 fdt_magic[] = {0xd0, 0x0d, 0xfe, 0xed};
 const u8 kernel_magic[] = {'A', 'R', 'M', 0x64};   // at 0x38
 const u8 cpio_magic[] = {'0', '7', '0', '7', '0'}; // '1' or '2' next
+const u8 macho_magic[] = {0xcf, 0xfa, 0xed, 0xfe};
 const u8 empty[] = {0, 0, 0, 0};
 
 struct kernel_header *kernel = NULL;
@@ -157,6 +159,9 @@ static void *load_one_payload(void *start, size_t size)
     } else if (!memcmp(p + 0x38, kernel_magic, sizeof kernel_magic)) {
         printf("Found a kernel at %p\n", p);
         return load_kernel(p, size);
+    } else if (!memcmp(p, macho_magic, sizeof macho_magic)) {
+        printf("Found a Mach-O image at %p\n", p);
+        return macho_load(p, size);
     } else if (!memcmp(p, empty, sizeof empty)) {
         printf("No more payloads at %p\n", p);
         return NULL;
@@ -172,6 +177,9 @@ int payload_run(void)
 
     while (p)
         p = load_one_payload(p, 0);
+
+    if (macho_boot)
+        return macho_boot();
 
     if (kernel && fdt) {
         smp_start_secondaries();
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ OBJECTS := \
 	heapblock.o \
 	iodev.o \
 	kboot.o \
+	macho.o \
 	main.o \
 	memory.o memory_asm.o \
 	payload.o \


src/macho.c
```diff
@@ -0,0 +1,84 @@
...
+#include "macho.h"
+#include "malloc.h"
+#include "memory.h"
+#include "utils.h"
+#include "xnuboot.h"
+
+#include "libfdt/libfdt.h"
+
+extern u64 boot_args_addr;
+extern void mmu_shutdown(void);
```
svenpeter42, line 12: this shouldn't be required since you include memory.h

src/payload.c
```diff
@@ -173,6 +178,9 @@ int payload_run(void)
     while (p)
         p = load_one_payload(p, 0);
 
+    if (macho_boot)
```
svenpeter42, line 181: this mixes an API for booting what I think is a raw macho payload with an API meant to boot a Linux kernel.

If we want to add support for this into m1n1 this needs more thoughts on where to put it and how to decide when we want to load a raw payload vs. a Linux kernel.

src/macho.h
```diff
@@ -0,0 +1,42 @@
...
+#define MACHO_COMMAND_UNIX_THREAD 0x05
+#define MACHO_COMMAND_SEGMENT_64  0x19
+
+struct macho_header {
+    u32 irrelevant[5];
+    u32 cmdsize;
+    u32 irrelevant2[2];
+};
+
+struct macho_command {
```
svenpeter42, line 17: I'd make these PACKED to be safe in case some compiler decides to introduce additional padding between the members

src/macho.c
```diff
@@ -0,0 +1,84 @@
...
+                        virtpc = dest + vmaddr - vmbase + pcoff;
+                    }
+                }
+            }
+        }
+        command = (void *)command + command->size;
+    }
+
+    macho_start_pc = virtpc;
+    macho_boot = macho_boot_impl;
```
svenpeter42, line 81: this is a little bit strange. you declare a global function pointer which will then only ever point to a single function.

Can't you use the return value?

svenpeter42 (commented):
The idea sounds useful for people who don't always have a proxy connected.

I think the implementation needs more thoughts though. Do you use this to chain load to a XNU kernel?

pipcet:
> I think the implementation needs more thoughts though.

I agree. Thanks for your comments!

> Do you use this to chain load to a XNU kernel?

The Corellium-based Linux kernel, currently.

src/payload.c
pipcet: > this mixes an API for booting what I think is a raw macho payload with an API meant to boot a Linux kernel.

I don't think payload.c is intended to be specific to Linux kernel payloads, but if it is, we can add another layer of indirection between `run_actions()` and `payload_run()`?

In my case, the macho payload is also a Linux kernel, plus the Corellium preloader.

> If we want to add support for this into m1n1 this needs more thoughts on where to put it and how to decide when we want to load a raw payload vs. a Linux kernel.

There's only ever one "real" payload, so I don't understand what would need to be decided.

src/macho.c
pipcet: > this is a little bit strange. you declare a global function pointer which will then only ever point to a single function.

Indeed, but a function pointer seems cleaner for future changes than having a global flag. I didn't want to go and change the existing Linux image code more than necessary, but it could be made to use the same function pointer.

> Can't you use the return value?

The return value of `macho_load()`? That needs to be `NULL` so we stop looking for additional payloads, doesn't it?

An alternative would be for `macho_load()` not to return at all, but I have a hunch it might be useful to include additional payloads.

pipcet:
> > Do you use this to chain load to a XNU kernel?
> 
> The Corellium-based Linux kernel, currently.

One thing I'd like to try, though, is to steal a CPU core from MacOS, then use that to run privileged code while MacOS is running. If it works, that should be easier than setting up a full hypervisor...

src/payload.c
svenpeter42: Sorry, my original comment was a little bit confusing.

Right now payload.c loads various payloads, prepares the FDT, spins up the other cores and jumps to a payload which has a linux-compatible header.

Now you want to introduce support for "raw" payloads that don't have a linux-kernel header and neither need the FDT nor the other cores nor support any other payloads so far.

I'm not sure we want to support this in the long term since we won't need the Corellium preloader anymore in the near future. But if we (read: marcan ;)) wants to support this we should think a little bit more where to put that code. I feel like it should be separate from the normal payload handling since it's doing something very different.

pipcet: > Sorry, my original comment was a little bit confusing.

No need to apologize. Again, thank you for your comments!

> Right now payload.c loads various payloads

Yes. That's what I want to do for macho payloads as well.

> prepares the FDT,

No, kboot.c does that.

> spins up the other cores

That should also be moved to kboot.c, I think.

> and jumps to a payload

It'd still do that.

> which has a linux-compatible header.

... but this requirement would be relaxed. In fact, it already is relaxed, since we support compressed payloads.

> Now you want to introduce support for "raw" payloads that don't have a linux-kernel header and neither need the FDT nor the other cores nor support any other payloads so far.

Yes. It's a different kind of payload, so it doesn't use kboot.c, but it's still a payload, so it uses payload.c. You're right that, right now, macho payloads cannot be combined with other payloads, so that part of the payload.c code isn't useful yet, but decompression should work...

> I'm not sure we want to support this in the long term since we won't need the Corellium preloader anymore in the near future. 

I think it would be impolite to go out of our way in order not to support booting the Corellium kernels.

(As for what we want in the long term, I'm afraid my ideas are a bit different. m1n1 is a good thing and possibly necessary to get us there, but I would much prefer to boot to a Linux kernel .macho image directly, load an initrd, and kexec() the real kernel from there, without using m1n1 in the ordinary boot process. In fact, I'd like to go the reverse route and prepare a kexec()-able m1n1 image that we can boot from Linux for experimentation and compatibility...)

> But if we (read: marcan ;)) wants to support this we should think a little bit more where to put that code. I feel like it should be separate from the normal payload handling since it's doing something very different.

I don't see why we need another layer of abstraction: there are payloads, handled by payload.c; macho images are one (handled by macho.c), Linux images are another (handled by kboot.c), and possibly booting MacOS or Windows would require a third type of payload (in the case of MacOS, we probably want to override the ADT (which would be another payload type) so we can keep some hardware for our nefarious purposes, but I'm not even sure we can boot to MacOS at all)).

Another type of payload which I think would be useful would be a simple if-then payload, so we can switch between payloads based, for example, on which USB port is connected or whether we're on AC power.

```
cat m1n1.macho if-no-usb-connected.payload linux.macho
```

Sorry this got a bit long: in summary, I think we should treat macho payloads as "normal" payloads, and move the Linux-Image-specific stuff to kboot.c. We should provide a way for people to use the Corellium kernel (which is very usable) or switch to m1n1 for development without going through the long-press-`bputil` process every time.

svenpeter42: > 
> > I'm not sure we want to support this in the long term since we won't need the Corellium preloader anymore in the near future.
> 
> I think it would be impolite to go out of our way in order not to support booting the Corellium kernels.

Not adding support for something because it makes the code more complex is hardly the same as going out of our way to not support booting these kernels.

> 
> Sorry this got a bit long: in summary, I think we should treat macho payloads as "normal" payloads, and move the Linux-Image-specific stuff to kboot.

Exactly! All I'm saying is that right now payload and kboot are rather strongly coupled and if we want to support different kinds of chain load sequences (e.g. with fdt/smp vs. just raw macho/elf/whatever file) we should actually separate this.

>We should provide a way for people to use the Corellium kernel (which is very usable) or switch to m1n1 for development without going through the long-press-`bputil` process every time.

Don't get me wrong, I completely agree that it's useable. It's just that we will make m1n1 more complex and potentially create more maintenance work in the future for something that might be obsolete soon. 
No one really knows what Corellium's plans are and for all we know they won't release another kernel.


I'd definitely wait for marcan to decide if he wants to support these kinds of payloads.

pipcet: > > > I'm not sure we want to support this in the long term since we won't need the Corellium preloader anymore in the near future.
> > I think it would be impolite to go out of our way in order not to support booting the Corellium kernels.
> Not adding support for something because it makes the code more complex is hardly the same as going out of our way to not support booting these kernels.

I'm sorry I worded that so strongly. This is hardly the right place to discuss project governance and the attitude toward Stan@Corellium's work.

> > Sorry this got a bit long: in summary, I think we should treat macho payloads as "normal" payloads, and move the Linux-Image-specific stuff to kboot.
> 
> Exactly! All I'm saying is that right now payload and kboot are rather strongly coupled and if we want to support different kinds of chain load sequences (e.g. with fdt/smp vs. just raw macho/elf/whatever file) we should actually separate this.

I'm still not sure what kind of separation you're proposing. Moving stuff from payload.c to kboot.c is hardly a big deal?

> > We should provide a way for people to use the Corellium kernel (which is very usable) or switch to m1n1 for development without going through the long-press-`bputil` process every time.
> Don't get me wrong, I completely agree that it's useable. It's just that we will make m1n1 more complex and potentially create more maintenance work in the future for something that might be obsolete soon.

I'm sorry if my code made it seem like macho is a complex format. It's not. The cost-benefit tradeoffs seem quite clear-cut to me.

> No one really knows what Corellium's plans are and for all we know they won't release another kernel.

"For all we know they won't" sounds like you have good reason to believe they won't. Is that the case? Anyway, even in that case, I think Stan's code is a much better starting point than starting over from scratch...

> I'd definitely wait for marcan to decide if he wants to support these kinds of payloads.

I'd definitely appreciate his comments or opinions.

marcan:
I'm not sure this is a feature we want in m1n1. If you want to dual-boot, the correct thing to do is to set up multiple operating systems as detailed in our quickstart guide; that is how the M1 boot process is intended to work, not "overloading" one OS with multiple OSes.

Loading Mach-Os like this would only work with m1n1 itself (rather useless) and the Corellium preloader (which I don't find particularly interesting); it can't load macOS because that requires more complex Mach-O command support, plus it will probably run into problems with RVBAR.

src/payload.c
svenpeter42: > > 
> > 
> > Exactly! All I'm saying is that right now payload and kboot are rather strongly coupled and if we want to support different kinds of chain load sequences (e.g. with fdt/smp vs. just raw macho/elf/whatever file) we should actually separate this.
> 
> I'm still not sure what kind of separation you're proposing. Moving stuff from payload.c to kboot.c is hardly a big deal?

No, I completely agree with you! Moving stuff from payload.c to kboot.c sounds like a good solution if we want to support these kind of payloads.

> 
> > > We should provide a way for people to use the Corellium kernel (which is very usable) or switch to m1n1 for development without going through the long-press-`bputil` process every time.
> > > Don't get me wrong, I completely agree that it's useable. It's just that we will make m1n1 more complex and potentially create more maintenance work in the future for something that might be obsolete soon.
> 
> I'm sorry if my code made it seem like macho is a complex format. It's not. The cost-benefit tradeoffs seem quite clear-cut to me.

The code implements a subset of macho. macho itself is much more complex which is what I'm afraid of.

> 
> > No one really knows what Corellium's plans are and for all we know they won't release another kernel.
> 
> "For all we know they won't" sounds like you have good reason to believe they won't. Is that the case? 

Wasn't meant to sound that way.

>Anyway, even in that case, I think Stan's code is a much better starting point than starting over from scratch...

Agreed.

marcan:
As for Corellium in general, considering they have had precisely zero interactions with the community, even though they have been repeatedly invited and CCed on e-mails, I do not think there is anything else for us to do here. The ball is in their court. If they want to collaborate, they need to speak up. As it stands, I am not willing to directly incorporate any of their code, as I have reasons to be concerned about its provenance and development methodology, and this could only be clarified by being open and transparent with the rest of the community; they have not done so.

pipcet:
> I'm not sure this is a feature we want in m1n1. If you want to dual-boot, the correct thing to do is to set up multiple operating systems as detailed in our quickstart guide;

Sorry, I must totally have missed that. I don't see anything in there that actually explains how to switch between OSes?

>  that is how the M1 boot process is intended to work, not "overloading" one OS with multiple OSes.

It's not "intended" to boot Linux at all, is it? Generally my experience has been that you less you rely on proprietary firmware which isn't supposed to work with Linux, the better...

pipcet:
> As for Corellium in general, considering they have had precisely zero interactions with the community, even though they have been repeatedly invited and CCed on e-mails, I do not think there is anything else for us to do here.

I was assuming that I wasn't the only one to hear back from them.

marcan:
It's intended to boot multiple OSes, just like e.g. EFI machines can, and each one can be whatever kernel you want. If you follow our [guide](https://github.com/AsahiLinux/docs/wiki/Developer-Quickstart), you end up with multiple "macOS" installs, one of which is Linux. You can switch from the boot picker by holding down the power button, or change the default by going into options and then going into startup options. Nothing stops you from doing two extra partitions like that, and making one use the Corellium kernel.

marcan:
Note that this built-in dual-boot feature isn't just cosmetic. It ties deeply into the M1's boot architecture. We will never support "dual-booting" macOS from m1n1 except for research purposes (tethered hypervisor) for this reason. It goes agains the secureboot design and is clearly not how things are intended to work. Keeping OSes separate allows us to have separate SEP contexts, separate security states, and avoids a million problems with Apple's tooling around updates and such that we will run into. It is a bad idea overall.

This is less of an issue if you're dual-booting Linux kernels, but that's what the u-boot and GRUB stages after m1n1 in production set-ups will be for. Which leaves the Corellium preloader as the only reason to load Mach-Os from m1n1. And sorry, I just really don't care about that.

pipcet:
Thanks for taking the time to explain. I don't think I could disagree more, but this is hardly the right place to discuss such questions.

Closing for now, let me know if you change your mind :-)

marcan:
If you disagree with the first part, I would be happy to go into further detail on the technical reasons why chainloading macOS as well as Linux from a single m1n1 install / boot policy is a positively terrible idea.

If you disagree with the second part, there isn't much more to discuss :-)

marcan:
Sure, but

> Being able to chain load the kernel would allow me to mutate the device tree as opposed to patching the kernel.

... that still means you need to 

> patch the kernel, bundle it together with the loaded kexts, reboot into 1TR, set the kernel, and reboot.

for every update, except instead of literally patching the kernel you stick m1n1 in front. What's the difference?

EricRabil:
The difference is I'm not going in and hex editing the kernel for a simple system feature (which is still a fun exercise), and have faster iteration times if I want to tinker with other parts of the device tree. It also eliminates the maintenance commitment to my patchfinder in the event they change the layout/logic of csr_bootstrap.

marcan:
If you want fast iteration times, just install m1n1 in proxy mode and use another host to upload whatever kernels and set whatever device tree properties you want :)

EricRabil:
That doesn't support the use case of a daily driver though. I've booted my Mac from the hypervisor before, but it isn't a practical way of sideloading my own code signing CA outside my office where I have a spare Mac.

marcan:
TBH, it sounds like you want to fork m1n1 and implement the required functionality yourself. My policy is basically that m1n1 tethered python scripts are a free-for-all and I'd happily take a script that does what you want, but for built-in functionality we only support things related to booting Asahi Linux or supporting the Python scripts/hypervisor functionality itself. I don't want m1n1 to grow into a giant kitchen sink; it's already getting complex enough as it is.

PR #53, opened 2021-05-06
guarded execution experiments

svenpeter42:


proxyclient/sprr_test_permissions.py
```diff
@@ -0,0 +1,386 @@
+#/usr/bin/env python3
+
+from setup import *
+import asm
+from contextlib import contextmanager
+
+p.smp_start_secondaries()
+
+class GuardedHeap:
```
marcan, line 9: I like this! Please put it in proxyutils.py :)

... but ...

proxyclient/sprr_test_permissions.py
```diff
@@ -0,0 +1,386 @@
...
+                mrs x12, {elr}
+                add x12, x12, #8
+                and x12, x12, x11
+                msr {elr}, x12
+
+                isb
+                {eret}
+        """.format(eret=eret, elr=elr, indicator=indicator))
+
+with guarded_heap() as heap:
```
marcan, line 143: ... why are you using it here? The utils heap is managed Python-side, so everything vanishes after the script exits, which is why currently almost nothing here cares about cleaning up allocations.

This context manager will probably be useful in the future, once we start e.g. using the hypervisor which will be a long-running script and may need to manage more complex allocation situations.

proxyclient/sprr_test_permissions.py
```diff
@@ -0,0 +1,386 @@
+#/usr/bin/env python3
```
marcan, line 1: 4-space indents for this file, please.

proxyclient/sprr_test_permissions.py
```diff
@@ -0,0 +1,386 @@
...
+                ldp x16, x17, [sp], #16
+                ldp x18, x19, [sp], #16
+                ldp x20, x21, [sp], #16
+                ldp x22, x23, [sp], #16
+                ldp x24, x25, [sp], #16
+                ldp x26, x27, [sp], #16
+                ldp x28, x29, [sp], #16
+                ldr x30, [sp], #16
+
+	        ret
```
marcan, line 303: Weird indentation?

src/cpu_regs.h
```diff
@@ -123,3 +123,66 @@
 
 #define SYS_IMP_APL_UPMSR sys_reg(3, 7, 15, 6, 4)
 #define UPMSR_IACT        (BIT(0))
+
+/* SPRR and GXF registers */
+#define SYS_SPRR_CONFIG_EL1       sys_reg(3, 6, 15, 1, 0)
```
marcan, line 128: Please follow the SYS_IMP_APL_ convention here (I know it's verbose, but it's what we're using in Linux, so let's keep it the same). sys_reg defines only, the bits don't need the prefix.

I care less about the Python side, so no need to stick IMP_APL_ in front of everything in the regs file. Perhaps just amend the generator script so it adds that prefix on request or where required.

src/cpu_regs.h
```diff
@@ -123,3 +123,66 @@
...
+
+#define SYS_GXF_ABORT_EL12 sys_reg(3, 6, 15, 15, 3)
+
+#define SYS_GXF_ENTER_EL12 sys_reg(3, 6, 15, 15, 2)
+
+#define SYS_SPRR_PERM_EL0 sys_reg(3, 6, 15, 1, 5)
+
+#define SYS_SPRR_PERM_EL1 sys_reg(3, 6, 15, 1, 6)
+
+#define SYS_TPIDR_GL1 sys_reg(3, 6, 15, 10, 1)
```
marcan, line 158: Personally, for related blocks of registers which definitely won't have bit defines (e.g. because they are just addresses or mirror non-impdef equivalents), I'd prefer if you group them in some reasonable way. Makes it more readable than a uniform list of registers with a single blank line after each.

src/exception.c
```diff
@@ -124,6 +167,10 @@ void print_regs(u64 *regs, int el12)
     u64 elr = el12 ? mrs(ELR_EL12) : mrs(ELR_EL1);
     u64 esr = el12 ? mrs(ESR_EL12) : mrs(ESR_EL1);
 
+    if (in_gl12()) {
```
marcan, line 170: These are the extra guarded mode bits, right? Only available in GLx (no ASPSR_EL1 etc)? Sounds good to do it like this, but maybe put the print right after SPSR_EL1 since it extends it.

marcan (changes requested):

proxyclient/sprr_test_permissions.py
svenpeter42: Because this code originally looked quite different and the guarded heap was a leftover I forgot to remove :)

proxyclient/sprr_test_permissions.py
svenpeter42: fixed.

proxyclient/sprr_test_permissions.py
svenpeter42: yup, fixed.

src/cpu_regs.h
svenpeter42: fixed.

src/exception.c
svenpeter42: Yes, that register only exists in guarded mode. I haven't found ASPSR_EL1 and I don't believe that it exists

src/cpu_regs.h
svenpeter42: I've added a flag to the generator script to add that IMP_APL prefix and fixed the defines

PR #56, opened 2021-05-14
usb/hv: add second CDC ACM interface for virtualized UART

jannau:

Second ACM interface is always announced but will be only used for the virtualized uart. Depends on using the other interface of the same USB device as proxy/hv connection. No support in `proxy/`.

USB device/interface handling is awkward and probably not using the right abstraction.
Read support not implemented.


marcan:
Tweaked a few things, I think this is good to merge :)

PR #58, opened 2021-05-16
Add address lookup for MMIO trace

jannau:


proxyclient/adt.py
```diff
@@ -246,17 +249,60 @@ def tostruct(self):
     def build(self):
         return ADTNodeStruct.build(self.tostruct())
 
+    def walk_tree(self):
+        yield self
+        for child in self:
+            for node in child.walk_tree():
```
marcan, line 255: `yield from`?

proxyclient/hv.py
```diff
@@ -39,6 +39,19 @@ class HOOK(IntEnum):
     "data" / Hex(Int64ul),
 )
 
+class RangedDict(dict):
```
marcan, line 42: I like the interface here (I was thinking of building something like this), but this does a linear search over the keys which is O(n), and it also doesn't handle overlapping ranges. Can we do this in O(log n) and handle arbitrary overlaps?

The algorithm I'm thinking about is like this: keep the regular dict stuff for the main storage, but then add a cache of *non-overlapping* intervals stored as a sorted list of (start, end, set(ranges)) tuples, initialized to [(0, 1<<64, set())]. On dict insertion, find the range of intervals that overlap the new range (if any) using the bisect module, and add the current range to all, splitting any that need to be split. Then lookup becomes a simple bisection of the interval set (O(log n)), and automatically gives you a set of matching ranges (you just need to check the interval end address to accept or reject the candidate range). Insertion would be O(log n) in the total interval count and O(n) in the number of overlapping intervals. If you want to support removal, you can just remove a range from the covered intervals; I don't think we need to bother with re-coalescing.

(There may be better ways of doing this, feel free to come up with something better :))

proxyclient/hv.py
```diff
@@ -151,7 +165,13 @@ def handle_mmiotrace(self, data):
         else:
             t = "R"
 
-        print(f"[0x{evt.pc:016x}] MMIO: {t} 0x{evt.addr:x} = 0x{evt.data:x}")
+        try:
+            dev, start, size = self.mmio_lockup[evt.addr]
```
marcan, line 169: And then, if you do the above, I think I'd rather just pre-populate mmio_lookup with all the ADT regs from the get go, instead of doing it lazily; this avoids having to do a whole ADT walk for every new range, we can just do it all in one walk.

proxyclient/adt.py
```diff
@@ -246,17 +249,60 @@ def tostruct(self):
...
+                except AttributeError:
+                    continue
+                if size > 0 and addr <= ptr and ptr < addr + size:
+                    results.append((node.name, addr, size))
+        if len(results) == 0:
+            return ('unknown', ptr & ~3, 4)
+        # sort by addr
+        s = sorted(results, key=itemgetter(1))
+        # return most specific match range
+        return sorted(s, key=itemgetter(2))[0]
```
marcan, line 277: And then we can just allow this to return all matches :)

proxyclient/hv.py
marcan: Also, this will be very useful later, as I plan to have "MMIO watcher" classes that can handle specific peripherals, and they need to be able to register themselves as watching certain MMIO ranges; I could use this same data structure for that, mapping ranges to watcher callbacks.

marcan: Also, since we'd be handling mutations and multiple results and such, at this point I'd drop the dict subclassing and just do it as a new class; all we really need are `.add(range, value)` and `.lookup(addr)` to begin with.

marcan:
To add context to the changes requested: I have a feeling mmiotrace is going to end up bottlenecked on python processing speed at the end of the day (since we have a big fat USB2 pipe), hence the focus on algorithmic efficiency.

proxyclient/adt.py
jannau: fixed

proxyclient/hv.py
jannau: replaced with `bisect` and sorted lists. address lookup is factor 30 faster. I also got rid of the dict as underlying storage. Storing range, device_id tuples works as well and is c. 10% faster.

proxyclient/hv.py
jannau: done

PR #59, opened 2021-05-16
proxy: Use Adler-32 for data checksum if possible

VinDuv:
This changes the checksum algorithm used for the proxy’s `MEMREAD` and `MEMWRITE` operations to use Adler-32: this algorithm is built-in on the Python side (and much faster than the manual checksum algorithm), and already included in m1n1 (part of `tinf`).
To ensure compatibility, the checksum algorithm is only used if both sides agree on using it; this is determined by feature flags that are exchanged during the initial `REQ_NOP` sent by `boostrap_port`:
* The checksum algorithm used for NOP messages is not modified by this feature so feature detection will work in all cases
* If an older version of m1n1 is used, it will ignore the feature flags in the NOP message and return 0 in the reply message, so no features will be enabled.
* If an older version of the proxy is used, it will send in its NOP message; m1n1 will disable all features and return 0 in the reply message.

### DIFF
--- a/proxyclient/proxy.py
+++ b/proxyclient/proxy.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # SPDX-License-Identifier: MIT
 
-import os, sys, struct, serial, time
+import os, sys, struct, serial, time, zlib
 from construct import *
 from utils import *
 from sysreg import *
@@ -75,6 +75,18 @@ class UartChecksumError(UartError):
 
 class UartRemoteError(UartError):
     pass
+
+class Feature(IntFlag):
+    DATA_CSUM_ADLER32 = 0x01  # Memory transfers use Adler-32 checksum
+
+    @classmethod
+    def get_all(cls):
+        return cls.DATA_CSUM_ADLER32
+
+    def __str__(self):
+        return ", ".join(feature.name for feature in self.__class__
+            if feature & self) or "<none>"
+
 
 class START(IntEnum):
     BOOT = 0
@@ -156,6 +168,7 @@ class UartInterface:
         self.tty_enable = True
         self.handlers = {}
         self.evt_handlers = {}
+        self.enabled_features = Feature(0)
 
     def checksum(self, data):
         sum = 0xDEADBEEF;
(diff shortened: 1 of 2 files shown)


jannau:
This is for 40 and 1000 byte payloads at least 50 times faster than the python or numpy code from #60

VinDuv:
Note that this PR does not affect MMIO tracing; to use Adler32 for this, the `REQ_EVENT` handler on the Python side needs to use `data_checksum` instead of `checksum`, and the `uartproxy_send_event` function on the C side needs to check if Adler32 is enabled and compute the checksum accordingly.
I didn’t try to implement it because I can’t test it currently; a slight difficulty is that `uartproxy_send_event` computes the checksum in two steps, but the `tinf_adler32` function can’t do that directly.

marcan:
How does this compare with #60, performance-wise? If we can get fast enough results with numpy then I'd rather keep the current checksum rather than introduce a feature flag.

Personally though, if we're introducing feature flags, I'd rather just set the checksum field to a magic value and forgo actual checksums. This should be enough to catch garbage/alignment issues (which are the primary cause of checksum failures in my experience), and with USB as the transport for anything performance-sensitive, checksums are largely redundant.

jannau:
For a stupid benchmark calling the checksum repeatedly on the same input for different data sizes:
```
checksum data size 16, iterations 3276800
adler32 imp: 0.46 s
numpy  code: 19.01 s
python code: 11.22 s
checksum data size 40, iterations 1310720
adler32 imp: 0.20 s
numpy  code: 7.64 s
python code: 10.69 s
checksum data size 132, iterations 397188
adler32 imp: 0.07 s
numpy  code: 2.32 s
python code: 10.35 s
checksum data size 512, iterations 102400
adler32 imp: 0.03 s
numpy  code: 0.65 s
python code: 10.23 s
checksum data size 16384, iterations 3200
adler32 imp: 0.02 s
numpy  code: 0.07 s
python code: 10.17 s
```

VinDuv:
I made a new PR (#61 ) that uses feature flags to disable checksums (and replace them with sentinel values). It should also work with MMIO traces. Closing this PR.

PR #60, opened 2021-05-16
proxy: use numpy for the checksum if possible

svenpeter42:
~twice as fast for small buffers (40 bytes), ~75 times faster for large buffers (4096 bytes) on my machine

### DIFF
--- a/proxyclient/proxy.py
+++ b/proxyclient/proxy.py
@@ -8,6 +8,35 @@ from sysreg import *
 from enum import IntEnum, IntFlag
 from serial.tools.miniterm import Miniterm
 
+try:
+    import numpy
+    from functools import lru_cache
+
+    numpy.seterr(over='ignore')
+
+    @lru_cache
+    def calc_checksum_powers(size):
+        powers = numpy.power(31337, numpy.arange(size-1, -1, -1, dtype=numpy.uint32))
+        base = (0xDEADBEEF * numpy.power(31337, size)) & 0xFFFFFFFF
+        return powers, base
+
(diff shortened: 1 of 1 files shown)


jannau:
numpy is unfortunately much slower for shorter messages. For 250k iterations over 40 byte (length of the mmiotrace packet) I get
```
python code: 1.73 s
numpy code: 3.86 s
```

svenpeter42:
Aw, what a pity. With some caching the 40 byte buffers are ~30% faster now on my machine. I feel like this is slowly moving into fun but unnecessary optimisation territory though ;)

```python
import numpy
from functools import lru_cache

@lru_cache
def checksum_get_base(l):
	return 0xDEADBEEF * numpy.power(31337, l)

@lru_cache
def checksum_get_powers(l):
	return numpy.power(31337, numpy.arange(l-1, -1, -1, dtype=numpy.uint32))

def np_checksum(data):
	data = numpy.frombuffer(data, dtype=numpy.uint8) ^ 0x5a
	powers = checksum_get_powers(len(data))
	base = checksum_get_base(len(data))
	csum = numpy.dot(data, powers)
	return ((csum + base) ^ 0xADDEDBAD) & 0xFFFFFFFF
```

marcan:
Looks good, but probably not worth it if we merge #61. Checksum overhead is negligible for UART, and I'd rather move towards dropping real checksums (and just in general fixing dumb problems with reentrancy/etc that were the cause of most checksum problems) when we're on USB, since the underlying protocol takes care of this for us.

PR #61, opened 2021-05-17
Add an option to disable checksumming if possible

VinDuv:
(based on feedback from previous PR #59 )

This allows disabling checksumming on memory reads, writes, and MMIO traces by setting the `DISABLE_CSUMS` environment variable.
To ensure compatibility, the checksum disabling is only used if both sides agree on using it; this is determined by feature flags that are exchanged during the initial `REQ_NOP` sent by `boostrap_port`:
* The checksum algorithm used for NOP messages is not modified by this feature so feature detection will work in all cases
* If an older version of m1n1 is used, it will ignore the feature flags in the NOP message and return 0 in the reply message, so no features will be enabled.
* If an older version of the proxy is used, it will send in its NOP message; m1n1 will disable all features and return 0 in the reply message.

### DIFF
--- a/src/uartproxy.c
+++ b/src/uartproxy.c
@@ -21,6 +21,7 @@ typedef struct {
             u64 size;
             u32 dchecksum;
         } mrequest;
+        u64 features;
     };
     u32 checksum;
 } UartRequest;
@@ -36,6 +37,7 @@ typedef struct {
             u32 dchecksum;
         } mreply;
         struct uartproxy_msg_start start;
+        u64 features;
     };
     u32 checksum;
     u32 _dummy; // Not transferred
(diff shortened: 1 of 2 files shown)


src/uartproxy.c
```diff
@@ -180,6 +196,9 @@ int uartproxy_run(struct uartproxy_msg_start *start)
 
         switch (request.type) {
             case REQ_NOP:
+                enabled_features = request.prequest.opcode & PROXY_FEAT_ALL;
```
marcan, line 199: Can you add a `features` member to the union, to avoid overloading the proxy request struct?

src/uartproxy.c
```diff
@@ -193,7 +212,7 @@ int uartproxy_run(struct uartproxy_msg_start *start)
                     break;
                 exc_count = 0;
                 exc_guard = GUARD_RETURN;
-                checksum_val = checksum((void *)request.mrequest.addr, request.mrequest.size);
+                checksum_val = data_checksum((void *)request.mrequest.addr, request.mrequest.size);
```
marcan, line 215: Ahh, but this won't do what I want. The goal of the sentinel is to catch data offset errors (e.g. bytes were added or removed), but that only works if the checksum comes *after* the data, which is not the case for data transfers, only for the commands and events.

Can you *add* a new sentinel field after the data to the data transfer ops? This does increase overhead slightly, but it is unlikely to matter given transfer overhead. With the new iodev_queue thing, this can be done without triggering unnecessary USB packet splitting overhead. You can use a different sentinel for this from the one used for checksum_val, so that if some weird alignment happens it won't falsely pass.

marcan (changes requested):

marcan:
I like redefining the iface-level NOP command as a feature exchange, it makes perfect sense :)

marcan:
Also, I'm thinking I want this to be always disabled for UART (checksums always enabled). One way to do it would be for the proxy side to always reject the feature if the current iodev is the UART one. Then the Python side can safely request it by default, and we probably don't even need the environment variable.

src/uartproxy.c
VinDuv: Makes sense, will fix.

src/uartproxy.c
VinDuv: I added the extra sentinel for both reads and writes. I’m not completely sure if [the way I flush queued writes](https://github.com/AsahiLinux/m1n1/commit/653a5497bbf239a9534ccb200fde72d22394f68d#diff-f84ef3ef6c6fcb6f5e938547f2a8f6cd000e68c6fd7a39639a40e19fbdd7362bR283) in the reply handler is correct, though.

VinDuv:
I replaced the `DISABLE_CSUMS` environment variable check with a `IODEV_UART` check on the M1’s side, as you suggested, so the checksum disabling is now automatic. That was easier to implement than I though it would be :-)

marcan (approved):

PR #62, opened 2021-05-18
Update README.md

MaddTheSane:
Fix formatting.
This makes it format better on GitHub.

### DIFF
--- a/font/README.md
+++ b/font/README.md
@@ -1,3 +1,4 @@
-These fonts have been generated with
+These fonts have been generated with:
+
 	makefont.sh 8 16 12 SourceCodePro-Bold.ttf font.bin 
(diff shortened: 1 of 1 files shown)


font/README.md
```diff
@@ -1,3 +1,4 @@
-These fonts have been generated with
+These fonts have been generated with:
+
 	makefont.sh 8 16 12 SourceCodePro-Bold.ttf font.bin
```
jannau, line 3: Please also add code clock via "```" for the two makefont.sh lines

MaddTheSane: The indentation alone makes it format as code.
If I add the code block (```) then it includes the indentation along with the code.

marcan:
Looks good, but the commit needs a signoff.

PR #65, opened 2021-05-29
add i2c and usb pd support and initialize the phy for the payload

svenpeter42:


alyssarosenzweig:
13:04 <@bloom> +int i2c_smbus_write32(i2c_dev_t *dev, u8 addr, u8 reg, u32 val)
13:04 <@bloom> do you not have memcpy in m1n1? :-p
13:05 <@bloom> memcpy(bfr, &val, 4)
13:05 <@bloom> +    u32 cmd_status = 0xffffffff; // just needs to be different from 0 and TPS_CMD_INVALID
13:05 <@bloom> use a do-while?

svenpeter42:
good point about the do/while loop. i've been writing too much python code lately ;)

regarding that memcpy: I deliberately wrote it out with shifts because it's easier to read for me that way and also makes the endianness explicit.

kettenis:
This would allow me to drop the code that sends SSPS from u-boot isn't it? That would be nice since that's currently done using an ugly hack that isn't easy to clean up.

svenpeter42:
Yep, with this there's no need to send the SSPS command for anything launched by m1n1 anymore unless the PD chip is somehow reset.

kettenis:
Verified that this does indeed make sending the SPSS from U-Boot redundant. @marcan, any chance of getting this merged? It looks good to me.

marcan:
Going to do a merge-round tonight!

marcan (approved):

PR #68, opened 2021-06-03
hv: shadow perf monitor IRQ mode and state

jannau:
The development Mac OS kernel panics if the PMCR0 sysreg is in
an unexpected state. To avoid that the hypervisor needs to shadow the
interrupt mode and interrupt active flag after it mask the PM FIQ.
Mac OS reads and writes frequently from PMCR0 and PMC 0/1 so handling
in m1n1 is preferred over Python.

Also disables PMU counters in the hypervisor.


leeehui:
@jannau 
sorry, I cannot figure out the overall picture, so currently m1n1 set pmc interrupt mode to FIQ, and handled in xnu? I wonder what is the problem currently?

jannau:
@leeehui,
xnu uses `PMC0` and `PMC1` and panics if `PMCR0` is not in the expected state.

marcan (approved):

PR #70, opened 2021-06-05
hv: add interrupt tracing

jannau:
Implemented by MMIO tracing of AIC's event register. Proposed by pipcet.

Would be nice if we could hook just the event register at offset 0x2004.

### DIFF
--- a/proxyclient/hv.py
+++ b/proxyclient/hv.py
@@ -30,6 +30,12 @@ EvtMMIOTrace = Struct(
     "data" / Hex(Int64ul),
 )
 
+EvtIRQTrace = Struct(
+    "flags" / Int32ul,
+    "type" / Hex(Int16ul),
+    "num" / Int16ul,
+)
+
 class HV_EVENT(IntEnum):
     HOOK_VM = 1
     VTIMER = 2
@@ -89,6 +95,9 @@ class HV:
         GXF_ENTER_EL1: GXF_ENTER_EL12,
     }
 
+    AIC_EVT_TYPE_HW = 1
+    IRQTRACE_IRQ = 1
+
     def __init__(self, iface, proxy, utils):
         self.iface = iface
         self.p = proxy
@@ -104,6 +113,7 @@ class HV:
         self._in_handler = False
         self._sigint_pending = False
         self.vm_hooks = []
+        self.interrupt_map = {}
 
     def unmap(self, ipa, size):
         assert self.p.hv_map(ipa, 0, size, 0) >= 0
@@ -128,6 +138,14 @@ class HV:
         index = len(self.vm_hooks)
         self.vm_hooks.append((read, write, ipa, kwargs))
         assert self.p.hv_map(ipa, (index << 2) | t, size, 1) >= 0
+
+    def trace_irq(self, device, num, count, flags):
+        for n in range(num, num + count):
+            if flags & self.IRQTRACE_IRQ:
+                self.interrupt_map[n] = device
+            else:
+                self.interrupt_map.pop(n, None)
+        assert self.p.hv_trace_irq(self.AIC_EVT_TYPE_HW, num, count, flags) > 0
 
     def addr(self, addr):
         unslid_addr = addr + self.sym_offset
@@ -170,6 +188,13 @@ class HV:
 
         print(f"[0x{evt.pc:016x}] MMIO: {t}.{1<<evt.flags.WIDTH:<2}{m} " +
               f"0x{evt.addr:x} ({dev}, offset {evt.addr - zone.start:#04x}) = 0x{evt.data:x}")
+
+    def handle_irqtrace(self, data):
+        evt = EvtIRQTrace.parse(data)
+
+        if evt.type == self.AIC_EVT_TYPE_HW and evt.flags & self.IRQTRACE_IRQ:
+            dev = self.interrupt_map[int(evt.num)]
+            print(f"IRQ: {dev}: {evt.num}")
 
     def handle_vm_hook(self, ctx):
         data = self.iface.readstruct(ctx.data, VMProxyHookData)
@@ -571,6 +596,7 @@ class HV:
         self.iface.set_handler(START.HV, HV_EVENT.VTIMER, self.handle_exception)
         self.iface.set_handler(START.HV, HV_EVENT.WDT_BARK, self.handle_bark)
         self.iface.set_event_handler(EVENT.MMIOTRACE, self.handle_mmiotrace)
+        self.iface.set_event_handler(EVENT.IRQTRACE, self.handle_irqtrace)
 
         #self.map_sw(0x2_00000000,
                     #0x2_00000000 | self.SPTE_TRACE_READ | self.SPTE_TRACE_WRITE,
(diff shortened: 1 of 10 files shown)


marcan:
Wait, why do we need all this code? Can't we just turn on mmiotrace reads of the AIC event register using the existing mmiotrace infrastructure? Python can then parse it into something useful like we'll do with all other mmiotrace stuff.

marcan:
Clearly I need to document the hypervisor :-)

jannau:
tracing the just event register in python was too slow. I was testing with just one of the USB-C ports disabled. This causes an IRQ storm since Mac OS can't mask the interrupts of the hidden port. #72 fixes that I would guess tracing in python will work.

marcan:
Oh, sorry, I missed that this lets you trace specific IRQs. Standard mmiotrace (note: not a synchronous python hook, just async mmiotrace) should be equivalent to using this code with all interrupts enabled, but of course being able to filter by IRQ is probably worth it.

src/hv_aic_events.c
```diff
@@ -0,0 +1,166 @@
...
+static bool check_hw_num(u16 num)
+{
+    u8 val = trace_hw_num[num / 8];
+    return !!(val & (1 << (num & 0x7)));
+}
+
+static void emit_irqtrace(u16 type, u16 num)
+{
+    struct hv_evt_irqtrace evt = {
+        .flags = 0,
```
marcan, line 43: Let's have a IRQTRACE_IRQ = 1 enum thing here for the flags, marking this as an IRQ event. This is considering we might want to trace mask/unmask operations from the guest in the future. Making this a bitmask keeps it consistent with the configuration function (see below).

src/hv_aic_events.c
```diff
@@ -0,0 +1,166 @@
...
+    };
+
+    hv_wdt_suspend();
+    uartproxy_send_event(EVT_IRQTRACE, &evt, sizeof(evt));
+    hv_wdt_resume();
+}
+
+static bool trace_aic_event(u64 addr, u64 *val, bool write, int width)
+{
+    if (write) {
```
marcan, line 55: Can you factor this passthrough out into a generic function in hv_vm.c, to be used by any similar tracers? In fact, it should be 3 functions: hv_pa_read, hv_pa_write, and hv_pa_rw (that takes the `bool write` arg and picks which one to call). The read/write ones can then be called in the `SPTE_MAP` cases in hv_vm.c and that keeps out duplication.

src/hv_aic_events.c
```diff
@@ -0,0 +1,166 @@
...
+            break;
+        default:
+            // ignore
+            break;
+    }
+
+    return true;
+}
+
+void hv_trace_aic_events(u32 type, u32 num, u32 active)
```
marcan, line 122: And then s/active/flags/, passing `IRQTRACE_IRQ` to enable tracing the IRQ firing.

Also, it might make sense to have a `u32 from, u32 count` instead of num, so that e.g. tracing all IRQs doesn't require hundreds of calls.

Let's also call this `hv_trace_irq` for consistency.

src/hv_aic_events.c
```diff
@@ -0,0 +1,166 @@
...
+#define AIC_EVENT_TYPE_HW   1
+#define AIC_EVENT_TYPE_IPI  4
+#define AIC_EVENT_IPI_OTHER 1
+#define AIC_EVENT_IPI_SELF  2
+
+#define AIC_MAX_HW_NUM 1024
+
+static u64 aic_base;
+
+static u8 trace_hw_num[AIC_MAX_HW_NUM / 8];
```
marcan, line 20: Since AIC works in terms of 32bit registers for masks/unmasks, let's also have the masks use u32 units. Doesn't matter much here, but for mask/unmask tracing (which would use other arrays that don't exist yet) that means we can use boolean ops to check if any interesting IRQ was masked/unmasked in one go.

src/hv_aic_events.c
```diff
@@ -0,0 +1,166 @@
...
+
+static void set_hw_num(u16 num)
+{
+    u8 *b = &trace_hw_num[num / 8];
+    *b |= 1 << (num & 0x7);
```
marcan, line 25: This is `BIT(num & 7)`. (or `& 31` when switching to u32), and at that point it's probably better to just do it in one line of code. Same for the ones below.

src/hv_aic_events.c
```diff
@@ -0,0 +1,166 @@
...
+static void clear_hw_num(u16 num)
+{
+    u8 *b = &trace_hw_num[num / 8];
+    *b &= ~(1 << (num & 0x7));
+}
+
+static bool check_hw_num(u16 num)
+{
+    u8 val = trace_hw_num[num / 8];
+    return !!(val & (1 << (num & 0x7)));
```
marcan, line 37: You don't need `!!` when the function already returns bool.

src/hv_aic_events.c
```diff
@@ -0,0 +1,166 @@
...
+            return;
+        }
+    }
+
+    static int hooked = 0;
+
+    if (aic_base && !hooked) {
+        hv_map_hook(aic_base, trace_aic_event, 0x4000);
+        hooked = 1;
+    }
```
marcan, line 165: Maybe complain if we can't find the AIC, possibly actually return an error code to the proxy?

marcan (changes requested):

src/hv_aic_events.c
jannau: done

src/hv_aic_events.c
jannau: all done I hope

proxyclient/hv.py
```diff
@@ -129,6 +139,13 @@ def map_hook(self, ipa, size, read=None, write=None, **kwargs):
         self.vm_hooks.append((read, write, ipa, kwargs))
         assert self.p.hv_map(ipa, (index << 2) | t, size, 1) >= 0
 
+    def trace_irq(self, device, num, count, flags):
+        if flags & self.IRQTRACE_IRQ:
+            self.interrupt_map[num] = device
+        else:
+            self.interrupt_map.pop(num)
```
marcan, line 146: This needs a guard in case the IRQ was already removed.

src/hv_aic.c
```diff
@@ -0,0 +1,119 @@
...
+
+    return true;
+}
+
+bool hv_trace_irq(u32 type, u32 num, u32 count, u32 flags)
+{
+    dprintf("hv_trace_irq type: %u start: %u num: %u flags: 0x%x\n", type, num, count, flags);
+    if (type == AIC_EVENT_TYPE_HW) {
+        if (num >= AIC_MAX_HW_NUM || count > AIC_MAX_HW_NUM - num) {
+            printf("hv: invalid IRQ range: (%u, %u)\n", num, num + count);
```
marcan, line 72: The convention in the other files is HV: in uppercase.

src/hv_aic.c
```diff
@@ -0,0 +1,119 @@
...
+            return false;
+        }
+
+        if (adt_get_reg(adt, adt_path, "reg", 0, &aic_base, NULL) < 0) {
+            printf("hv: Error getting AIC base address.\n");
+            return false;
+        }
+    }
+
+    static int hooked = 0;
```
marcan, line 106: bool?

marcan (changes requested):

marcan:
Looks good, just a few minor nits. Thanks! I'll fix them when I get to the merge later if you don't beat me to it :)

proxyclient/hv.py
jannau: Was intended to be `pop(num, None)`, missed to add all interrupts to mapping dictionary. both fixed

marcan (approved):

marcan:
Fixed an unimportant nit and merged :)

PR #72, opened 2021-06-09
usb: disable tps6598x interrupts

jannau:
Restore the interrupt masks on transition to the next state. The masks
are not restored on the USB-C port used by the hypervisor. This prevents
an interrupt storm in the guest when the other USB-C port is exposed to
the guest. The tps6598x interrupt line is shared amoung both ports.

### DIFF
--- a/proxyclient/m1n1/hv.py
+++ b/proxyclient/m1n1/hv.py
@@ -584,6 +584,12 @@ class HV:
         self.tba = self.u.ba.copy()
         self.device_addr_tbl = self.adt.build_addr_lookup()
 
+        # disable unused USB iodev early so interrupts can be reenabled in hv_init()
+        for iodev in (IODEV.USB0, IODEV.USB1):
+            if iodev != self.iodev:
+                print(f"Disable iodev {iodev!s}")
+                self.p.iodev_set_usage(iodev, 0)
+
         print("Initializing hypervisor over iodev %s" % self.iodev)
         self.p.hv_init()
 
--- a/proxyclient/m1n1/hv.py
+++ b/proxyclient/m1n1/hv.py
@@ -717,19 +717,19 @@ class HV:
                         del self.adt[name]
                     except KeyError:
                         pass
-            for name in ("/cpus/cpu1",
-                         "/cpus/cpu2",
-                         "/cpus/cpu3",
-                         "/cpus/cpu4",
-                         "/cpus/cpu5",
-                         "/cpus/cpu6",
-                         "/cpus/cpu7",
-                        ):
-                print(f"Removing ADT node {name}")
-                try:
-                    del self.adt[name]
-                except KeyError:
-                    pass
+        for name in ("/cpus/cpu1",
+                     "/cpus/cpu2",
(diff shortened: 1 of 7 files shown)


marcan:
Can we just unconditionally do this on init? If m1n1 isn't using/handling these IRQs it should just disable them by default.

jannau:
Mac OS seems to rely on the default config or iboot for the pd controller's interrupt config. When the interrupts are disabled in m1n1 the port is dead.
We could declare that HV needs port 0 and only disable interrupts it.

marcan:
We've run into things like this in the past, so I think I'm going to rework how we do subsystem shutdowns a bit. I'll discriminate between "shutdown for kboot" and "shutdown for chainload"; the latter tries to restore state to whatever iBoot leaves to make macOS (or another m1n1) happy, while the former leaves things in the sanest state for Linux. Let me think for a bit how I want to do this :)

jannau:
I think the sanest state for chainload and linux boot does not differ in this case. Leaving the interrupts in the default config looks sane to me. The Problem is the hv guest. It wouldn't be a problem there either if we were not using one port for the hv or if both ports were not sharing the same interrupt line.

Kept now mostly within m1n1. The only required proxy change is that the other USB iodev gets disabled in `hv.init()` so m1n1 knows in `hv_init` for which port it should restore the interrupt mask.

Also in this PR some minor python hv fixes and removal of the hiding of the second usb-c port.

proxyclient/m1n1/hv.py
```diff
@@ -677,7 +683,7 @@ def rh(base, off, width):
             print(f"R {base:x}+{off:x}:{width} = 0x{data:x} -> 0x{ret:x}")
             return ret
 
-        for addr in (0x23b700420, 0x23d280098, 0x23d280088, 0x23d280090):
+        for addr in (0x23b700420, 0x23d280098, 0x23d280088):
```
marcan, line 686: Does this cover the case where the HV is using both ports? I've only tested with port 0 so far, I imagine the problem addresses are different for each.

marcan:
I don't like the idea of leaving IRQs enabled for Linux (the sane state for IRQ hardware is leaving stuff masked), but we can shave that yak later. Other than the comment this looks good, feel free to mark it ready.

proxyclient/m1n1/hv.py
jannau: I haven't noticed a difference after removing `0x23d280090` and still removing both ports from the guest's ADT.

It is however required when running running Mac OS as guest using HV on port 1. In that case it's necessary to hook `(0x23b700448, 0x23d2800a0, 0x23d280090)` though. I'll add that to this PR.

jannau:
Interrupts are now only restored for the guest and for chainloading (not via `chainload.py -c` though)
Writes to the tps6598x were not reliable. I've both failures to clear IntMask and to restore it to the boot values. Apparently "solved" by adding `udelay(120)` after i2c writes.
HV works now on both USB-C ports and the other port is usable in Mac OS.

PR #73, opened 2021-06-09
usb_dwc3: shutdown the DART as well when shutting down usb

svenpeter42:


kettenis:
This one would be helpful for u-boot as it means the existing dwc3 driver just works.

svenpeter42:
linux as well once a small bugfix (https://lore.kernel.org/linux-usb/YL3ve%2FrxTlIzErlX@infradead.org/T/#mfdcca059365fe9a5aff59279cf1a7ed9daa797db) reaches upstream

PR #75, opened 2021-06-10
run_guest.py: fix usage without '-m' or '-c'

jannau:


marcan:
Huh, I guess that action does not set the default value. Fixed a different way in e833e3c.

PR #76, opened 2021-06-11
kboot: Provide MAC addresses in device tree

kettenis:
Read the MAC addresses from the ADT and store them in
"local-mac-address" properties on the relevant nodes in the FDT.
This relies on appropriate aliases in the provided FDT, which
matches how U-Boot provides MAC addresses on embedded targets.

### DIFF
--- a/src/kboot.c
+++ b/src/kboot.c
@@ -244,6 +244,41 @@ static int dt_set_cpus(void)
     return 0;
 }
 
+static const char *aliases[] = {
+    "bluetooth0",
+    "ethernet0",
+    "wifi0",
+};
+
+static int dt_set_mac_addresses(void)
+{
+    int anode = adt_path_offset(adt, "/chosen");
+
+    if (anode < 0)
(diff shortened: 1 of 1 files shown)


kettenis:
To add a little bit more context:
The ADT defines the following properties under /chosen
- mac-address-bluetooth0
- mac-address-ethernet0
- mac-address-wifi0
and the code expects aliases with the following names under /aliases
- bluetooth0
- ethernet0
- wifi0
Having ethernetN aliases is already standard practice and U-Boot uses those on embedded platforms to provide local-mac-address properties in the device tree passed to the OS, so I think it makes sense for m1n1 to do something similar using the ADT.  We can't do this in U-Boot since U-Boot never sees the ADT.
Having separate wifiN and bluetoothN aliases instead of just ethernetN aliases makes life easier.
For now only N=0 is implemented. As long as Apple doesn't release hardware with multiple Ethernet ports it doesn't really make sense to make things more complicated.

PR #77, opened 2021-06-11
Add support for device specific mmiotrace handlers

jannau:
Adds a handler for "/arm-io/gpio" as example.


alyssarosenzweig:
> +        # ignore moisy SPI NOR chip select?

Typo

jannau:
> Typo

fixed, thanks

marcan:
I have a different idea of how to do the tracers, but the prep commits and general idea are good so I'll merge this as is; we can port the GPIO example to what I have in mind later :)

PR #81, opened 2021-06-26
Help cmd

amworsley:
Add a help command which is very useful for beginners to see what commands are available

shell.py or run_guest.py gain the help command that will list one line for
each command. Either from a doc string of the function when present or the fully qualified
name of the command for easy reference to the source code.
Some of the commands have doc strings added and more can be updated as I understand
how they work.

### DIFF
--- a/proxyclient/m1n1/proxy.py
+++ b/proxyclient/m1n1/proxy.py
@@ -80,6 +80,20 @@ ExcInfo = Struct(
     "sp_phys" / Int64ul,
     "data" / Int64ul,
 )
+# Sends 56+ byte Commands and Expects 36 Byte Responses
+# Commands are format <I48sI
+#   4 byte command, 48 byte null padded data + 4 byte checksum 
+# Responses are of the format: struct format <Ii24sI
+#   4byte Response , 4 byte status, 24 byte string,  4 byte Checksum
+#    Response must start 0xff55aaXX where XX distiguishes between them
+#    In little endian mode these numbers as listed as REQ_* constants
+# defined under UartInterface
+#
+#  Event Response REQ_EVENT passed to registered Event Handler
+#  Boot Response REQ_BOOT passed to handle_boot() which may
+#       pass to a matching registered handler based on reason, code values
+#  If the status is ST_OK returns the data field to caller
+#     Otherwise reports a remote Error
 
 class UartInterface(Reloadable):
     REQ_NOP = 0x00AA55FF
@@ -426,6 +440,8 @@ REGION_RWX_EL0 = 0x8000000000
 REGION_RW_EL0 = 0x9000000000
 REGION_RX_EL1 = 0xa000000000
 
+# Uses UartInterface.proxyreq() to send requests to M1N1 and process
+# reponses sent back.
 class M1N1Proxy(Reloadable):
     S_OK = 0
     S_BADCMD = -1
--- a/proxyclient/m1n1/proxy.py
+++ b/proxyclient/m1n1/proxy.py
@@ -685,102 +685,107 @@ class M1N1Proxy(Reloadable):
         self.request(self.P_REBOOT, no_reply=True)
 
     def write64(self, addr, data):
-        '''write64(addr, data) - write 8 byte value to given address'''
+        '''write 8 byte value to given address'''
         if addr & 7:
             raise AlignmentError()
         self.request(self.P_WRITE64, addr, data)
     def write32(self, addr, data):
-        '''write32(addr, data)  - write 4 byte value to given address'''
+        '''write 4 byte value to given address'''
         if addr & 3:
             raise AlignmentError()
         self.request(self.P_WRITE32, addr, data)
     def write16(self, addr, data):
-        '''write16(addr, data) - write 2 byte value to given address'''
+        '''write 2 byte value to given address'''
         if addr & 1:
             raise AlignmentError()
         self.request(self.P_WRITE16, addr, data)
     def write8(self, addr, data):
-        '''write8(addr, data) - write 1 byte value to given address'''
+        '''write 1 byte value to given address'''
         self.request(self.P_WRITE8, addr, data)
 
     def read64(self, addr):
-        '''read64(addr) - return 8 byte value from given address'''
+        '''return 8 byte value from given address'''
         if addr & 7:
             raise AlignmentError()
         return self.request(self.P_READ64, addr)
     def read32(self, addr):
-        '''read32(addr) - return 4 byte value given address'''
+        '''return 4 byte value given address'''
         if addr & 3:
             raise AlignmentError()
         return self.request(self.P_READ32, addr)
     def read16(self, addr):
-        '''read16(addr) - return 2 byte value from given address'''
+        '''return 2 byte value from given address'''
         if addr & 1:
             raise AlignmentError()
         return self.request(self.P_READ16, addr)
     def read8(self, addr):
-        '''read8(addr) - return 1 byte value from given address'''
+        '''return 1 byte value from given address'''
         return self.request(self.P_READ8, addr)
 
     def set64(self, addr, data):
-        '''set64(addr, data) - Or 64 bit value of data into memory at addr and return result'''
+        '''Or 64 bit value of data into memory at addr and return result'''
         if addr & 7:
             raise AlignmentError()
         self.request(self.P_SET64, addr, data)
     def set32(self, addr, data):
-        '''set32(addr, data) - Or 32 bit value of data into memory at addr and return result'''
+        '''Or 32 bit value of data into memory at addr and return result'''
         if addr & 3:
             raise AlignmentError()
         self.request(self.P_SET32, addr, data)
     def set16(self, addr, data):
-        '''set16(addr, data) - Or 16 bit value of data into memory at addr and return result'''
+        '''Or 16 bit value of data into memory at addr and return result'''
         if addr & 1:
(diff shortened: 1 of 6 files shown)


proxyclient/m1n1/proxyutils.py
```diff
@@ -86,17 +86,22 @@ def __init__(self, p, heap_size=1024 * 1024 * 1024):
         }
 
     def read(self, addr, width):
+        '''read(addr, width) - do a width read from addr and return it
+        width can be 8,16,21,64 or 132'''
```
marcan, line 90: s/132/128/

Also please add spaces after commas for style.

proxyclient/m1n1/proxyutils.py
```diff
@@ -86,17 +86,22 @@ def __init__(self, p, heap_size=1024 * 1024 * 1024):
...
         val = self._read[width](addr)
         if self.proxy.get_exc_count():
             raise ProxyError("Exception occurred")
         return val
 
     def write(self, addr, data, width):
+        '''write(addr, data, width) - do a width write of data to addr
+        width can be 8,16,21,64 or 132'''
```
marcan, line 98: Same as above.

proxyclient/m1n1/proxy.py
```diff
@@ -669,78 +685,102 @@ def reboot(self):
         self.request(self.P_REBOOT, no_reply=True)
 
     def write64(self, addr, data):
+        '''Addr - write 8 bytes to given address'''
```
marcan, line 688: Why not use the same syntax as the other commands, with the function name? i.e. `write64(addr, data) - ...`.

proxyclient/m1n1/proxyutils.py
```diff
@@ -86,17 +86,22 @@ def __init__(self, p, heap_size=1024 * 1024 * 1024):
...
         self._write[width](addr, data)
         if self.proxy.get_exc_count():
             raise ProxyError("Exception occurred")
 
     def mrs(self, reg, *, silent=False, call=None):
+        '''mrs(reg) - read system register reg'''
```
marcan, line 104: Perhaps mention that this takes a bunch of different sysreg styles as parsed by `sysreg_parse`: a tuple representing the encoding, a string identifier in "sN_N_cN_cN_N" form, or a register name if in the database.

proxyclient/m1n1/proxyutils.py
```diff
@@ -185,6 +191,8 @@ def push_adt(self):
         self.iface.writemem(adt_base, self.adt_data)
 
     def disassemble_at(self, start, size, pc=None):
+        '''Start, len, [pc] - disassemble len bytes of memory from start
```
marcan, line 194: Same as above, why not show the function call syntax?

proxyclient/m1n1/shell.py
```diff
@@ -82,11 +110,29 @@ def display(val):
...
         for attr in dir(sysreg):
             locals[attr] = getattr(sysreg, attr)
 
+        for obj_name in locals.keys():
+            obj = locals.get(obj_name)
+            if obj is None:
+                continue
+            if callable(obj) and not isinstance(obj, property):
+                desc = locals[obj_name].__doc__
+                if not desc:
```
marcan, line 125: What's this `re` stuff doing, exactly? If you want to get the class name and such, there are easier ways to do that without resorting to string parsing :)

marcan (changes requested):

proxyclient/m1n1/proxyutils.py
amworsley: Ok - I note there are a few other, unrelated, instances found by : git grep -e '[^, ],[^, ]'

proxyclient/m1n1/shell.py
amworsley: Please point me in the right direction. I thought there would be cleaner methods but I couldn't see any so a pointer would be very helpful.

proxyclient/m1n1/proxyutils.py
amworsley: Addressed in new commit

proxyclient/m1n1/proxyutils.py
amworsley: Addressed in new commit

proxyclient/m1n1/proxy.py
amworsley: Fixed in new update

proxyclient/m1n1/proxyutils.py
amworsley: Addressed in latest commit

proxyclient/m1n1/proxyutils.py
amworsley: Fixed in latest commit

amworsley:
Should I merge all the changes into one commit for clarity?

amworsley:
Did a rebase and force push to the latest changes

proxyclient/m1n1/shell.py
ludwigschubert: @amworsley Maybe [lib/inspect](https://docs.python.org/3/library/inspect.html) can be helpful? I see it has a way to check if something is a function — `inspect.isfunction()` — and a way to access it's qualified name — `obj. __qualname__ `. Hope that's the right direction!

amworsley: That looks fantastic, exactly the sort of thing I need.

Thanks very much Ludwig, I will have another go later today.

I had been working on another branch adding in some commands to dump the page tables and physical memory this item has blocked me from submitting
a new PR for it. So thanks again

Andrew

amworsley: Ok - switched to using the inspect class to get the Signature / command names and split the commands into categories so that it is much less cluttered.

See below for an example output:

```
..........................................................................................
Copying SEPFW (0x750000 bytes)...
Copying TrustCache (0x60000 bytes)...
Adjusting addresses in ADT...
Setting up bootargs at 0x81c0fc000...
Entering hypervisor shell. Type `start` to start the guest.
>>> help
Note: To display a category's commands quote the name e.g. help('HV')
List of top level commands:
TraceMode  : TraceMode(value, names=None, *, module=None, qualname=None, type=None, start=1) - 
HV         : HV category - 50 subcommands
map_hw     : map_hw(ipa, pa, size) - map IPA (Intermediate Physical Address) to actual PA
h          : h(number, /) - Return the hexadecimal representation of an integer.
UartInterface:
             UartInterface category - 17 subcommands
M1N1Proxy  : M1N1Proxy category - 108 subcommands
clear16    : clear16(addr, data) - Clear bits in 16 bit memory at address addr that are set
clear32    : clear32(addr, data) - Clear bits in 32 bit memory at address addr that are set
clear64    : clear64(addr, data) - Clear bits in 64 bit memory at address addr that are set
clear8     : clear8(addr, data) - Clear bits in 8 bit memory at addr that are set in data
hv_translate:
             hv_translate(addr, s1=False, w=False) - Translate virtual address
mask32     : mask32(addr, clear, set) - Clear bits in 32 bit memory at address addr that are
mask64     : mask64(addr, clear, set) - Clear bits in 64 bit memory at address addr that are
mask8      : mask8(addr, clear, set) - Clear bits in 1 byte memory at addr that are set
mset64ask16:
             mset64ask16(addr, clear, set) - Clear select bits in 16 bit memory addr that are set
read16     : read16(addr) - return 2 byte value from given address
read32     : read32(addr) - return 4 byte value given address
read64     : read64(addr) - return 8 byte value from given address
read8      : read8(addr) - return 1 byte value from given address
set16      : set16(addr, data) - Or 16 bit value of data into memory at addr and return result
set32      : set32(addr, data) - Or 32 bit value of data into memory at addr and return result
set64      : set64(addr, data) - Or 64 bit value of data into memory at addr and return result
set8       : set8(addr, data) - Or byte value of data into memory at addr and return result
write16    : write16(addr, data) - write 2 byte value to given address
write32    : write32(addr, data) - write 4 byte value to given address
write64    : write64(addr, data) - write 8 byte value to given address
write8     : write8(addr, data) - write 1 byte value to given address
ProxyUtils : ProxyUtils category - 16 subcommands
disassemble_at:
             disassemble_at(start, size, pc=None) - disassemble len bytes of memory from start
dump_mem   : dump_mem(ba=34359738368, n=48) - dump_mem(ba=0x8_0000_0000, len=48) - hex dump n bytes memory from ba
dump_pt    : dump_pt(arg=None, debug=None) -  dump_pt(debug=level) - dump translation tables
mrs        : mrs(reg, *, silent=False, call=None) - read system register reg
msr        : msr(reg, val, *, silent=False, call=None) - Write val to system register reg
read       : read(addr, width) - do a width read from addr and return it
write      : write(addr, data, width) - do a width write of data to addr
DABORT_DFSC:
             DABORT_DFSC(value, names=None, *, module=None, qualname=None, type=None, start=1) - An enumeration.
ESR_EC     : ESR_EC(value, names=None, *, module=None, qualname=None, type=None, start=1) - An enumeration.
Enum       : Enum(value, names=None, *, module=None, qualname=None, type=None, start=1) - 
IntEnum    : IntEnum(value, names=None, *, module=None, qualname=None, type=None, start=1) - Enum where members are also (and must be) ints
MSR_DIR    : MSR_DIR(value, names=None, *, module=None, qualname=None, type=None, start=1) - An enumeration.
SPSR_M     : SPSR_M(value, names=None, *, module=None, qualname=None, type=None, start=1) - An enumeration.
```

PR #84, opened 2021-07-01
fix linux boot issues on macOS

svenpeter42:
without this macOS complains about the USB device that goes missing
once m1n1 hands control to Linux

### DIFF
--- a/src/uartproxy.c
+++ b/src/uartproxy.c
@@ -284,6 +284,7 @@ int uartproxy_run(struct uartproxy_msg_start *start)
 
         // Flush all queued data
         iodev_write(iodev, NULL, 0);
+        iodev_flush(iodev);
     }
 
     return ret;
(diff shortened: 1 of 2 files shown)


marcan:
This doesn't look right; m1n1 should be sending the reply. Do we have a race where m1n1 does not flush the buffer properly? If so we should fix that (probably worth doing `iodev_flush(uartproxy_iodev);` after `iodev_write(...)` at the end of the proxy loop, which shouldn't hurt performance since obviously proxy stuff is synchronous anyway so we might as well wait for the data to go out). If that doesn't fix it then we have a race and the macOS driver does not return buffered data after a device disconnect, so, uh, add a 100ms udelay before USB shutdown in `main.c`? :/

svenpeter42:
turns out that this actually had two root causes:
1. the flush was indeed missing
2. the usb phy was re-initialized which confused mac os because the usb device reappeared

marcan:
I guess we're still relying on a race here (Linux will *eventually* reinit USB), but I guess that's more than enough time for the reply to make it out to python, and in the meantime since m1n1 leaves the USB hardware in limbo macOS won't detect a disconnect. Good enough for now. Might want to explicitly nuke USB (with a bit of a delay) in the future though :)

PR #88, opened 2021-07-15
fix typo of "arm-trusted-firmware" in README.md

goshhhy:
force-pushed to add DCO signoff. if signoff under a pseudonym is not acceptable, please close without merging.


marcan:
Pseudonyms are fine here, just be aware that they may not be for other projects.

PR #89, opened 2021-07-26
startup: Prettyify video depth dump

alyssarosenzweig:
Use the right unit. for rgb10x2, this prints "30bpp" instead of "0x1E".

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>

### DIFF
--- a/src/startup.c
+++ b/src/startup.c
@@ -59,7 +59,7 @@ void dump_boot_args(struct boot_args *ba)
     printf("    stride:     0x%lx\n", ba->video.stride);
     printf("    width:      %lu\n", ba->video.width);
     printf("    height:     %lu\n", ba->video.height);
-    printf("    depth:      0x%lx\n", ba->video.depth);
+    printf("    depth:      %lubpp\n", ba->video.depth);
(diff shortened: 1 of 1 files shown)


src/startup.c
```diff
@@ -59,7 +59,7 @@ void dump_boot_args(struct boot_args *ba)
     printf("    stride:     0x%lx\n", ba->video.stride);
     printf("    width:      %lu\n", ba->video.width);
     printf("    height:     %lu\n", ba->video.height);
-    printf("    depth:      0x%lx\n", ba->video.depth);
+    printf("    depth:      %lubpp\n", ba->video.depth);
```
pipcet, line 62: Is it intentional that this now prints "65566bpp" for my MacBook with the retina flag set? That made me think I had a broken m1n1 version, and so I wasted some time trying to track it down...

marcan: Should be fixed in 19b16f77d.

pipcet: Thanks, can confirm that fixes the issue!

PR #96, opened 2021-08-14
proxyclient: removed unused code

TophEvich:
Little clean-up of unused proxyclient code.
Code used/referenced by commented out sections were taken into account and left behind to be used by it at a later point.

Signed-off-by: TophEvich <84676511+TophEvich@users.noreply.github.com>

### DIFF
--- a/proxyclient/m1n1/hv.py
+++ b/proxyclient/m1n1/hv.py
@@ -186,9 +186,9 @@ class HV(Reloadable):
     def map_hook(self, ipa, size, read=None, write=None, **kwargs):
         index = len(self.vm_hooks)
         self.vm_hooks.append((read, write, ipa, kwargs))
-        self.map_hook_idx(ipa, size, index, read is not None, write is not None, kwargs)
+        self.map_hook_idx(ipa, size, index, read is not None, write is not None)
 
-    def map_hook_idx(self, ipa, size, index, read=False, write=False, kwargs={}):
+    def map_hook_idx(self, ipa, size, index, read=False, write=False):
         if read:
             if write:
                 t = self.SPTE_PROXY_HOOK_RW
@@ -213,7 +213,7 @@ class HV(Reloadable):
         if len(self.interrupt_map):
             self.add_tracer(zone, "AIC_IRQ", TraceMode.RESERVED)
         else:
-            self.del_tracer(zone, "AIC_IRQ", TraceMode.RESERVED)
+            self.del_tracer(zone, "AIC_IRQ")
 
         assert self.p.hv_trace_irq(self.AIC_EVT_TYPE_HW, num, count, flags) > 0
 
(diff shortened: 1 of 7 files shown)


cristim:
How did you detect this code? Did you use any linter for it? 

If so we should probably introduce a lint check in the PR workflow

TophEvich:
Sorry I have almost no python experience and am a python ecosystem novice. I just wanted to have a look at the code and used PyCharm for easy browsing. So I didn't set out to lint and I wouldn't know the pros and cons of each of the python linters that are out there.
So it was PyCharms internal "linter" which you can't use in CI. Also I don't know how much it would help, I think the nature of the project has quite a lot of code that might be hard for linters and as a result you'd have a lot of false positives (bad signal-to-noise ratio).
I only committed the parts where I was quite confident that the PyCharm was correct and also looked for commented out sections as they might be used in debugging sessions and so on.

Many of the errors/warnings/notices I ignored because it looked to me that it was meta programming/project layout related false positives.

proxyclient/m1n1/trace/__init__.py
```diff
@@ -32,9 +32,9 @@ def read(self, addr, width):
             self.cache[addr] = data
             return data
         else:
-            return self.read_cached(addr, width)
+            return self.read_cached(addr)
 
-    def read_cached(self, addr, width):
```
marcan, line 37: Let's keep this one orthogonal with read().

proxyclient/m1n1/hv.py
```diff
@@ -698,10 +697,10 @@ def handle_sync(self, ctx):
             return self.handle_hvc(ctx)
 
         if ctx.esr.EC == ESR_EC.SSTEP_LOWER:
-            return self.handle_step(ctx)
```
marcan, line 701: Let's keep these orthogonal since ctx might be useful in the future.

proxyclient/m1n1/hv.py
```diff
@@ -213,7 +213,7 @@ def trace_irq(self, device, num, count, flags):
         if len(self.interrupt_map):
             self.add_tracer(zone, "AIC_IRQ", TraceMode.RESERVED)
         else:
-            self.del_tracer(zone, "AIC_IRQ", TraceMode.RESERVED)
+            self.del_tracer(zone, "AIC_IRQ")
```
marcan, line 216: Ah, this one's an actual bug :)

proxyclient/m1n1/adt.py
```diff
@@ -88,7 +88,7 @@ def parse_prop(node, path, name, v):
 
     return t, v
 
-def build_prop(path, name, v, t=None):
+def build_prop(name, v, t=None):
```
marcan, line 91: This is here because we might have path-dependent build rules in the future.

marcan (changes requested):

marcan:
The cleanup of unused variables looks good, but for the function args, a lot of the unused ones are there because they make sense and might become used in the future. I'd rather keep those.

TophEvich:
Rebased on latest main branch and put back in the sections you wanted because they are placeholder or too keep method signature symmetry in their respective "family".

marcan (approved):

PR #99, opened 2021-08-24
Avoid reporting bytes in the RX buffer when it's empty.

pipcet:
I believe this fix, or something like it, is necessary to avoid spurious NUL bytes being received over the vUART by the Linux driver.

(Unfortunately, I'm having a hard time testing the vUART right now. I'm using vanilla m1n1 on a machine that has never booted anything but vanilla m1n1 and macOS, and am booting straight from iBoot...)

Anyway, it's a pretty obvious fix :-)

### DIFF
--- a/src/hv_vuart.c
+++ b/src/hv_vuart.c
@@ -24,6 +24,7 @@ static void update_irq(void)
     utrstat |= UTRSTAT_TXBE | UTRSTAT_TXE;
     utrstat &= ~UTRSTAT_RXD;
 
+    ufstat = 0;
     if ((rx_queued = iodev_can_read(vuart_iodev))) {
         utrstat |= UTRSTAT_RXD;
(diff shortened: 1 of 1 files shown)


pipcet:
Just as a data point:

without this patch, I get a data abort and a reboot if I `python proxyclient/tools/run_guest.py m1n1.macho`, both when `m1n1.macho` is vanilla m1n1 and when it is produced with the patch.

if I chainload the `m1n1.macho` with this patch first, everything subsequently works as expected.

marcan:
The patch looks good for Linux. But there is no way this affects m1n1, since we don't read UFSTAT at all there. I think you're confusing yourself with something completely unrelated.

pipcet:
You're right, of course, thank you. I'll be looking for that other bug if and when it reappears :-)

PR #100, opened 2021-08-24
Cosmetic fixes, avoid relying on UTRSTAT_RXD == 1

pipcet:
Tiny readability fixes. If you don't want this kind of fix, please let me know, but at least this hunk:

```
-    return read32(UART_BASE + UTRSTAT) & UTRSTAT_RXD;
+    return (read32(UART_BASE + UTRSTAT) & UTRSTAT_RXD) ? 1 : 0;
```

seems important to me to make the code more obviously correct.

### DIFF
--- a/src/iodev.c
+++ b/src/iodev.c
@@ -33,7 +33,7 @@ size_t con_rp[IODEV_MAX];
 ssize_t iodev_can_read(iodev_id_t id)
 {
     if (!iodevs[id]->ops->can_read)
-        return false;
+        return 0;
 
     return iodevs[id]->ops->can_read(iodevs[id]->opaque);
 }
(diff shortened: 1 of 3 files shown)


marcan:
Can you prefix your commit description with `hv_vuart: `? I did it for your other PR, but please try to keep with the commit message style. It makes it much easier to look for relevant commits.

pipcet:
Sure, sorry, and sorry for the dumb question, but the prefixes are "logical", then, and not directly related to where the modified code lives? Because this PR doesn't touch the hv_uart code, of course.

marcan:
Sorry, that was a brain fart. I meant `uart: `. `hv_vuart` was for the other PR.

marcan:
Actually, I see you touched more things. In that case the commit description is not correct, since it only mentions the uart change. You should write something that encompasses all the changes.

PR #103, opened 2021-08-30
pcie: Only initialize ports that are actually present

kettenis:
The macbook pro and macbook air only have a single PCIe port

### DIFF
--- a/src/pcie.c
+++ b/src/pcie.c
@@ -185,6 +185,9 @@ int pcie_init(void)
 
         sprintf(bridge, "/arm-io/apcie/pci-bridge%d", port);
 
+        if (adt_path_offset(adt, bridge) < 0)
+            continue;
+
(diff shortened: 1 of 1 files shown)


alyssarosenzweig:
Looks sane ... I'm curious how that works though? I guess there's no Ethernet port, but do  they not separate wifi/eth from thunderbolt on the macbooks?

kettenis:
There are dedicated PCIe controllers for the Thunderbolt ports (both on the mini and the macbook air/pro). Those aren't touched by m1n1 at the moment. This is all about the PCIe controller used for the integrated devices (wifi/usb/ethernet).

alyssarosenzweig:
Oops, my bad. What I should have asked -- is there no separation between
USB type-A and WiFi  on the macbooks?

..Wait, are the macbooks type-C only now?

kettenis:
They are!

alyssarosenzweig:
Oh boy. My apologies!

PR #104, opened 2021-08-31
main: Fix compiler error

kettenis:
Older versions of GCC throw a "initializer element is not contstant"
error.  Fix this.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -24,7 +24,7 @@
 
 struct vector_args next_stage;
 
-const char *const version_tag = "##m1n1_ver##" BUILD_TAG;
+const char version_tag[] = "##m1n1_ver##" BUILD_TAG;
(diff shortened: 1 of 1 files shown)


kettenis:
This is the error message as I see it with

```
aarch64-linux-gnu-gcc (Linaro GCC 7.4-2019.02) 7.4.1 20181213
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

which is the aarch64 cross-compiler that is currently used by the OpenBSD ports system:

```
src/main.c:28:34: error: initializer element is not constant
 const char *const m1n1_version = version_tag + 12;
                                  ^~~~~~~~~~~
```

PR #110, opened 2021-09-21
Bit of audio

povik:
This is the code from a paste I posted earlier (https://gist.github.com/povik/66f69f539905bcde6f4cacbae1d24ad6) after some additional care.

(I will be happy to add sign offs once it's specified somewhere what those sign offs mean. :) )

### DIFF
--- a/proxyclient/experiments/speaker_amp.py
+++ b/proxyclient/experiments/speaker_amp.py
@@ -1,0 +1,236 @@
+# speaker_amp.py -- play audio through the embedded speaker on Mac mini
+#
+# sample usage with sox:
+#
+#   sox INPUT_FILE -t raw -r 48000 -c 1 -e signed-int -b 32 -L - gain -63 | python3 ./speaker_amp.py
+#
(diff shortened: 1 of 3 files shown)


povik:
If the sign-offs mean this then here they are: https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin

marcan:
Indeed we use the same DCO as Linux. We should probably add that to the README...

marcan:
Thanks! :)

I'm curious, are you interested in looking into writing a Linux driver for this? Audio has been on my list for right after the GPU gets on track, but with this much already figured out making an ASoC driver should be fairly easy. I'm happy to take care of it, but if you're interested feel absolutely free to go ahead too! It would be really cool to have basic audio working in the following month or two.

povik:
I am interested! I do not know how much time I will have, but I would definitely like to give it a try. Although if it happens that you or someone else steps in and gets it working in Linux before me, I wouldn't mind at all.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #113, opened 2021-10-02
Add smc.py script to enable radios

alyssarosenzweig:
Add a basic SMC client and a script in experiments/ using it to enable
the radios. Running this before booting Linux will make the network
controller devices show up in lspci, even if Linux lacks an SMC driver.

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>

### DIFF
--- a/proxyclient/experiments/smc.py
+++ b/proxyclient/experiments/smc.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
+import sys, pathlib
+import time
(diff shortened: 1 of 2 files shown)


alyssarosenzweig:
The use case is to investigate the Wi-Fi situation without sorting out SMC GPIO on the Linux side and wiring that into the PCIe probe path and ... doing that properly is nontrivial, and doing it with hacks would muck up the kernel tree too much. So just add a script to do it in m1n1 for now.

PR #119, opened 2021-11-06
usb: support more than 2 usb-c ports

jannau:
Adds support for the 3rd USB-C on the 2021 Macbook Pro.
Tested on Macbook Pro 14" and Mac Mini.

### DIFF
--- a/proxyclient/m1n1/hv.py
+++ b/proxyclient/m1n1/hv.py
@@ -983,8 +983,8 @@ class HV(Reloadable):
         self.print_tracer = trace.PrintTracer(self, self.device_addr_tbl)
 
         # disable unused USB iodev early so interrupts can be reenabled in hv_init()
-        for iodev in (IODEV.USB0, IODEV.USB1):
-            if iodev != self.iodev:
+        for iodev in IODEV:
+            if iodev >= IODEV.USB0 and iodev != self.iodev:
                 print(f"Disable iodev {iodev!s}")
                 self.p.iodev_set_usage(iodev, 0)
 
@@ -1086,10 +1086,7 @@ class HV(Reloadable):
             self.log(f"PMGR R {base:x}+{off:x}:{width} = 0x{data:x} -> 0x{ret:x}")
             return ret
 
-        if self.iodev == IODEV.USB0:
-            atc = "ATC0_USB"
-        elif self.iodev == IODEV.USB1:
-            atc = "ATC1_USB"
+        atc = f"ATC{self.iodev - IODEV.USB0}_USB"
 
         hook_devs = ["UART0", atc]
 
@@ -1166,8 +1163,8 @@ class HV(Reloadable):
         soc_name = "Virtual " + self.adt["product"].product_soc_name + " on m1n1 hypervisor"
         self.adt["product"].product_soc_name = soc_name
 
-        if self.iodev in (IODEV.USB0, IODEV.USB1):
-            idx = int(str(self.iodev)[-1])
+        if self.iodev >= IODEV.USB0:
+            idx = self.iodev - IODEV.USB0
             for prefix in ("/arm-io/dart-usb%d",
                            "/arm-io/atc-phy%d",
                            "/arm-io/usb-drd%d",
--- a/proxyclient/m1n1/hv.py
+++ b/proxyclient/m1n1/hv.py
@@ -1068,7 +1068,7 @@ class HV(Reloadable):
 
         zone = irange(base, 0x4000)
         irq = node.interrupts[0]
-        self.p.hv_map_vuart(base, irq, getattr(IODEV, self.iodev.name + "_SEC"))
+        self.p.hv_map_vuart(base, irq, self.iodev)
         self.add_tracer(zone, "VUART", TraceMode.RESERVED)
 
     def map_essential(self):
--- a/proxyclient/m1n1/proxy.py
+++ b/proxyclient/m1n1/proxy.py
@@ -83,7 +83,7 @@ ExcInfo = Struct(
 )
 # Sends 56+ byte Commands and Expects 36 Byte Responses
 # Commands are format <I48sI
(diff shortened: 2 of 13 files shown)


marcan:
This looks good, but let's be a bit more ambitious. Can you make it support up to 8 ports (which is the maximum a 2x M1 Max system could support)? That might involve refactoring the iodev stuff a bit to dynamically allocate/populate the structures for USB, instead of having static definitions for each port.

jannau:
Done, I've keep static iodev IDs and the full array of iodev pointer to keep the code simpler. I got rid of the secondary usb iodevs. They are not really needed since we will only use a single one.

proxyclient/m1n1/hv.py
```diff
@@ -1086,10 +1086,7 @@ def rh(base, off, width):
             self.log(f"PMGR R {base:x}+{off:x}:{width} = 0x{data:x} -> 0x{ret:x}")
             return ret
 
-        if self.iodev == IODEV.USB0:
-            atc = "ATC0_USB"
-        elif self.iodev == IODEV.USB1:
-            atc = "ATC1_USB"
+        atc = "ATC{}_USB".format(int(str(self.iodev)[-1]))
```
marcan, line 1089: Let's use an f-string here, since that's the style the rest of the code uses. Also, it probably makes sense to use the numeric value of the iodev and subtract USB0; that avoids the string munging and will trivially scale to >10 ports.

proxyclient/m1n1/proxy.py
```diff
@@ -425,6 +425,12 @@ class IODEV(IntEnum):
     FB = 1
     USB0 = 2
     USB1 = 3
+    USB2 = 4
+    USB3 = 5
+    USB4 = 6
+    USB5 = 7
+    USB6 = 8
+    USB7 = 9
```
marcan, line 433: We probably want to list the VUART here too. Can we move that one numerically before USB0? That way its value won't change as we add USB ports.

src/iodev.h
```diff
@@ -11,6 +11,12 @@ typedef enum _iodev_id_t {
     IODEV_FB,
     IODEV_USB0,
     IODEV_USB1,
+    IODEV_USB2,
```
marcan, line 14: And then if we make USB IODEVs the last in the list, we don't even have to actually list them all in the enum - just USB0 and MAX, with everything in between implied.

(Though we probably still want to list them in the python just so IntEnum doesn't complain about missing enum values)

Then USB_INSTANCES can be defined here (perhaps as USB_IODEV_COUNT instead) and usb.c can use that instead.

src/iodev.h
```diff
@@ -11,6 +11,12 @@ typedef enum _iodev_id_t {
...
+    IODEV_USB3,
+    IODEV_USB4,
+    IODEV_USB5,
+    IODEV_USB6,
+    IODEV_USB7,
     IODEV_USB_MAX,
```
marcan, line 20: I'd prefer it if MAX is inclusive, i.e. equal to USB7, not to one after, given the name.

src/usb.c
```diff
@@ -22,66 +23,70 @@ struct usb_drd_regs {
...
+#define FMT_ATC_PATH         "/arm-io/atc-phy%u"
+#define FMT_DRD_PATH         "/arm-io/usb-drd%u"
+#define FMT_HPM_PATH         "/arm-io/i2c0/hpmBusManager/hpm%u"
 
 static tps6598x_irq_state_t tps6598x_irq_state[USB_INSTANCES];
 static bool usb_is_initialized = false;
 struct usb_drd_regs usb_regs[USB_INSTANCES];
 
-static dart_dev_t *usb_dart_init(const char *path, const char *mapper_path)
+static int snprintf(char *buffer, size_t size, const char *fmt, ...)
```
marcan, line 36: This should go in utils.h/c. Though we already have sprintf (which is unsafe, but okay-ish for these use cases; feel free to replace the other uses with snprintf though while you're at it :))

src/usb.c
```diff
@@ -22,66 +23,70 @@ struct usb_drd_regs {
...
+{
+    va_list args;
+    int len;
+
+    va_start(args, fmt);
+    len = vsnprintf(buffer, size, fmt, args);
+    va_end(args);
+    if (len >= 0 && (size_t)len >= size)
+        printf("usb: Error praparing ADT path '%s'", fmt);
```
marcan, line 45: Typo, but also, this doesn't really belong in snprintf does it? :)

src/usb.c
```diff
@@ -22,66 +23,70 @@ struct usb_drd_regs {
...
+
+    return len;
+}
+
+static dart_dev_t *usb_dart_init(u32 idx)
 {
     int mapper_offset;
+    char path[sizeof(FMT_DART_MAPPER_PATH)];
```
marcan, line 53: Might be good adding a comment that this is good up to 100 instances, since that isn't immediately obvious unless you think about the width of the format specifier.

marcan (changes requested):
Looks good! Just some minor nits.

proxyclient/m1n1/hv.py
jannau: done and replaced another instance of the same string munging

proxyclient/m1n1/proxy.py
jannau: we don't really interact with VUART iodev from python but done

src/iodev.h
jannau: with `IODEV_USB_VUART` moved from the end IODEV_MAX is enough

src/usb.c
jannau: done, I missed the `sprintf` otherwise I would have used that

src/usb.c
jannau: comment and `#error` added

marcan (approved):

PR #123, opened 2021-11-14
sync dts with upcoming linux-5.17-rc3

jannau:
Syncs the included dts to include pcie and pinctrl support to be relased as linux 5.16-rc1.
Also adds dts files for all M1 Mac devices. This includes support for the 2020 MackBook Air and Pro and both 2021 24-inch iMac models. Changes are based on Apple's Monterey ADTs. I have only tested `t8103-j274.dts`.

I plan to send `arm64: dts: apple: add dts files for all M1 models` for kernel inclusion once the added models are tested. It passes `make dtbs_check` except for complains about missing required properties for `ethernet0: pci@0,0`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ OBJECTS := \
 	wdt.o \
 	$(MINILZLIB_OBJECTS) $(TINF_OBJECTS) $(DLMALLOC_OBJECTS) $(LIBFDT_OBJECTS)
 
-DTS := t8103-j274.dts
+DTS := $(patsubst dts/%,%,$(wildcard dts/*.dts))
 
 BUILD_OBJS := $(patsubst %,build/%,$(OBJECTS))
 DTBS := $(patsubst %.dts,build/dtb/%.dtb,$(DTS))
--- a/dts/dt-bindings/gpio/gpio.h
+++ b/dts/dt-bindings/gpio/gpio.h
@@ -1,0 +1,42 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/*
+ * This header provides constants for most GPIO bindings.
+ *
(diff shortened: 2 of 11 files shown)


svenpeter42:
As mentioned on IRC: I think syncing with what's in the kernel is awesome and thanks for taking care of that but I wouldn't add any changes to the device trees that haven't been reviewed on the mailing list yet.
No need to wait for the kernel merge window imho but they should be ready to go into linux-next before we merge them into m1n1.

While that's not important yet I'm a bit afraid that we might otherwise end up with DTs with minor issues on people's machines which might then break backwards- oder forwards-compatibility.

jannau:
I wanted to use it as staging ground before sending the change to the mailing list. I've removed the commit it can be dealt with in a second pull request.

jannau:
Patches are now based on a rewritten https://git.kernel.org/pub/scm/linux/kernel/git/devicetree/devicetree-rebasing.git tree.
Tree is rewritten with `git filter-repo --paths-from-file ../m1n1-apple-dts.lst` with following contents:
```
# List of files and dirs to keep
src/arm64/apple
include/dt-bindings/gpio/gpio.h
include/dt-bindings/interrupt-controller/apple-aic.h
include/dt-bindings/interrupt-controller/irq.h
include/dt-bindings/pinctrl/apple.h

# move according to m1n1 structure
src/arm64/apple==>dts
include/dt-bindings==>dts/dt-bindings
```
This requires manual work when an additional dt-bindings header is used. The header has to be added to the list and manually added to m1n1. It however ensures that changes to `include/dt-bindings` are included.

marcan:
Since we're now using U-Boot as the canonical DTS provider, we can close this.

PR #125, opened 2021-11-20
kboot: adjust max-link-speed for 10GE Mac Mini

jannau:
Modifies just "max-link-speed" although the template ADT for J274 has
different values for "pci-max-payload-size" and "pci-aspm-default".
The "lan-10gb" sub node has an additional property
"function-pcie_port_control".

### DIFF
--- a/src/kboot.c
+++ b/src/kboot.c
@@ -286,6 +286,52 @@ static int dt_set_mac_addresses(void)
     return 0;
 }
 
+/* adjust PCI port properties, Required for adjusting max-link-speed
+ * for the 10 gigabit ehternet Mac Mini variant.
+ */
+static int dt_adjust_pcie_port_properties(void)
+{
+    int anode = adt_path_offset(adt, "/arm-io/apcie");
+
+    if (anode < 0)
+        bail("ADT: /arm-io/apcie not found\n");
+
+    ADT_FOREACH_CHILD(adt, anode)
+    {
+        u32 port_num;
+        if (ADT_GETPROP(adt, anode, "apcie-port", &port_num) < 0)
+            continue;
+
+        // restrict to 100 ports, matches the numbering in the FDT and ought ti be enough
+        if (port_num > 99)
+            continue;
+
+        u32 max_link_speed;
+        if (ADT_GETPROP(adt, anode, "maximum-link-speed", &max_link_speed) < 0)
+            continue;
+
+        if (max_link_speed < 1) {
+            printf("ADT: unexpected maximum-link-speed %u for port %u\n", max_link_speed, port_num);
+            continue;
+        }
+
+        char port_name[16];
+        snprintf(port_name, sizeof(port_name), "port%02u", port_num);
+
+        const char *path = fdt_get_alias(dt, port_name);
+        if (path == NULL)
+            continue;
+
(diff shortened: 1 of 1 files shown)


jannau:
A simpler alternative to this m1n1 change would be to increase "max-link-speed" unconditionally in the dts. The gigabit ethernet Mac Mini model works with that change.

src/kboot.c
```diff
@@ -286,6 +286,52 @@ static int dt_set_mac_addresses(void)
...
+    if (anode < 0)
+        bail("ADT: /arm-io/apcie not found\n");
+
+    ADT_FOREACH_CHILD(adt, anode)
+    {
+        const char *anode_name = adt_get_name(adt, anode);
+        if (strncmp(anode_name, "pci-bridge", 10) || strlen(anode_name) != 11)
+            continue;
+
+        char port_num = anode_name[10];
```
marcan, line 305: Probably better to use the `apcie-port` prop here, instead of parsing the node name.

src/kboot.c
```diff
@@ -286,6 +286,52 @@ static int dt_set_mac_addresses(void)
...
+        if (ADT_GETPROP(adt, anode, "maximum-link-speed", &max_link_speed) < 0)
+            continue;
+
+        if (max_link_speed < 1 || max_link_speed > 16) {
+            printf("ADT: unexpected maximum-link-speed %u for %s\n", max_link_speed, anode_name);
+            continue;
+        }
+
+        char port_name[] = "port0_";
+        port_name[5] = port_num;
```
marcan, line 319: Undefined behavior here; this string is no longer NULL terminated. Might as well just use `snprintf()`, which is easier to read anyway.

src/kboot.c
```diff
@@ -286,6 +286,52 @@ static int dt_set_mac_addresses(void)
...
+
+        const char *path = fdt_get_alias(dt, port_name);
```
marcan, line 321: How does this work? We don't actually have aliases (in `/aliases`) for the PCIe ports, as far as I know.

marcan (changes requested):
I'm confused as to how this was tested, since the Linux aPCIE driver currently ignores max-link-speed entirely as far as I can tell. I don't understand how this fixed a real problem for you.

I think that if the 1GE version doesn't have issues down-negotiating from Gen4, then for now it's probably easier to just let it try up to that unconditionally, so we wouldn't need this code. However, we should certainly keep this branch around in case it becomes more useful in the future :)

src/kboot.c
jannau: if I'm not mistaken it replaces the '_' in "port0_" and the string is still NULL terminated. Port numbers use two digits. I agree that `snprintf` would be more readable.

src/kboot.c
jannau: agreed, I'll update the change

jannau:
I tested only the device tree rewriting by using a dtb with `max-link-speed` != 1 which m1n1 corrected to 1 from the 1GE ADT. I assumed that the link speed training was implemented since the PCIe ports already used different values.

jannau:
Closing since it is unnecessary since m1n1 sets the link speed based on the ADT since abee2ecebc4f7c71ede304dfc185f007944b64f5

PR #130, opened 2021-11-24
Various audio-related changes

povik:
These are some smaller changes that I have accumulated, with the addition of code for driving the NCO on top.


proxyclient/m1n1/hw/nco.py
```diff
@@ -0,0 +1,88 @@
+__all__ = ["NCO"]
```
marcan, line 1: This needs an SPDX license header

marcan (changes requested):

PR #134, opened 2021-12-15
Consistency for (c)

mmhj:
Could also use the symbol: &copy;

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,26 +66,26 @@ Supported compression formats:
 
 m1n1 is licensed under the MIT license, as included in the [LICENSE](LICENSE) file.
 
-* Copyright (C) 2021 The Asahi Linux contributors
+* Copyright (c) 2021 The Asahi Linux contributors
 
 Please see the Git history for authorship information.
 
 Portions of m1n1 are based on mini:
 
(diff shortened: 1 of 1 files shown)


mmhj:
Could possibly implement `&copy;` instead.

marcan:
Changing other people's copyright lines is generally a bad idea. We should change the Asahi Linux Contributors one to this format, though:

`Copyright The Asahi Linux Contributors`

As that is what we ought to be using for the kernel, and it is what the [Linux Foundation](https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/) recommends.

mmhj:
> Changing other people's copyright lines is generally a bad idea. We should change the Asahi Linux Contributors one to this format, though:
> 
> `Copyright The Asahi Linux Contributors`
> 
> As that is what we ought to be using for the kernel, and it is what the [Linux Foundation](https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/) recommends.

Fair enough.
I've changed only the `Copyright The Asahi Linux Contributors` in this new commit.

However, I still think there are some inconsistencies in regards to the dates at least.
Some of the license declarations should at some point be updated to reflect the correct dates.

PR #136, opened 2021-12-20
kboot: Provide antenna SKU in device tree

kettenis:
Read the antenna SKU from the ADT and store it in a
"apple,antenna-sku" property on the relevant node in the FDT.


kettenis:
Not sure if I remembered the property name correctly. And it might still change during review of the Linux code I guess.

PR #138, opened 2022-01-02
Add spmi experiment to read rtc clock

jfbortolotti:
This is a small experiment to read RTC time from Sera PMU using SPMI bus.

Signed-off-by: Jean-Francois Bortolotti <jeff@borto.fr>

### DIFF
--- a/proxyclient/experiments/spmi.py
+++ b/proxyclient/experiments/spmi.py
@@ -1,0 +1,87 @@
+import datetime
+
+# RTC is available from SERA PMU slave id/address 0xF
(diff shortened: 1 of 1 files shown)


marcan:
There's an SPMI class these days [here](https://github.com/AsahiLinux/m1n1/blob/main/proxyclient/m1n1/hw/spmi.py); any chance you could rewrite this on top of that, or close it if no longer relevalt?

Thanks :)

jfbortolotti:
Not sure it is really useful anymore as now RTC is supported over SMC and lina implemented a more comprehensive SPMI class.
Closing it.

PR #139, opened 2022-01-02
hv_vm: fix an alignment problem in hv_map

Zzzec:
A trivial fix since the input data are already aligned in most of cases, so the problem won't lead to something like crash. But BIT() should be the correct choose in ALIGN_UP/DOWN.


marcan:
Nice catch. I think the original code worked fine in all cases in practice (as long as everything is aligned to 2 bytes, which is guaranteed by an earlier check) by coincidence, but obviously BIT() is what was intended here.

PR #142, opened 2022-01-08
Replace (C) with Unicode Copyright symbol "©"

DestroyerXyz:


svenpeter42:
thanks for the contribution but please see the discussion in https://github.com/AsahiLinux/m1n1/pull/134#issuecomment-994635547

> Changing other people's copyright lines is generally a bad idea. We should change the Asahi Linux Contributors one to this format, though:
> 
> `Copyright The Asahi Linux Contributors`
> 
> As that is what we ought to be using for the kernel, and it is what the [Linux Foundation](https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/) recommends.

DestroyerXyz:
Yeah, that makes sense but most of the notices in the README aren't in that format?

marcan:
Those are other people's notices; we only get to pick how we format our own notice.

PR #143, opened 2022-01-10
ASC, SART, RTKit and NVMe Stub

svenpeter42:
the iova.c/h stuff is only very lightly tested. I originally thought I'd need that for NVMe for some reason.
SARTv3 (or anything else that requires the Pro/Max) isn't tested at all.

The NVMe part right now is just a stub that boots the co-processor to verify that the rest works correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ OBJECTS := \
 	main.o \
 	mcc.o \
 	memory.o memory_asm.o \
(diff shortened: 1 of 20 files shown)


marcan:
Looks good on first glance, let's merge it and I'll hack on it for the EPIC/SMC stuff. Thanks! :)

PR #144, opened 2022-01-16
display: Retry dcp_ib_get_hpd() for up to 2 seconds

jannau:
It takes 5 tries to detect my 1920x1080@60 displays (accumulated delay 400ms). 2 seconds selected to have 5 times margin of error and a not too annoying delay when no display is connected.
I hope we can decrease delay after we have seen how long the display detection usually takes.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -7,6 +7,9 @@
 #include "utils.h"
 #include "xnuboot.h"
 
+#define DISPLAY_STATUS_DELAY   100
+#define DISPLAY_STATUS_RETRIES 20
+
 #define COMPARE(a, b)                                                                              \
     if ((a) > (b)) {                                                                               \
         *best = modes[i];                                                                          \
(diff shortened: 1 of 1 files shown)


src/display.c
```diff
@@ -80,18 +80,30 @@ int display_configure(void)
...
+        } else if (hpd && timing_cnt && color_cnt) {
+            ret = 0;
+            printf("display: try:%d connected:%d timing_cnt:%d color_cnt:%d\n", retries, hpd,
+                   timing_cnt, color_cnt);
+            break;
+        }
+        if (retries < 20)
+            mdelay(100);
+    }
+    if (ret < 0) {
```
marcan, line 102: Shouldn't this also bail if hpd == 0? That's the display not connected case. Right now, as far as I can tell that will fall out of the loop and be treated as success.

marcan (changes requested):

src/display.c
jannau: you're right, I'll push a fix

PR #149, opened 2022-01-17
kboot: Add u-boot,dm-pre-reloc properties

kettenis:
These properties are needed such that U-Boot can enable the
serial port in its early pre-relocation phase.

### DIFF
--- a/src/kboot.c
+++ b/src/kboot.c
@@ -327,6 +327,59 @@ static int dt_set_wifi(void)
     return 0;
 }
 
+static void dt_set_uboot_dm_preloc(int node)
+{
+    // Tell U-Boot to bind this node early
+    fdt_setprop_empty(dt, node, "u-boot,dm-pre-reloc");
+
+    // Make sure the power domains are bound early as well
+    int pds_size;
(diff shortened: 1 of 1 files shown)


src/kboot.c
```diff
@@ -327,6 +327,49 @@ static int dt_set_wifi(void)
...
 
+static void dt_set_uboot_dm_preloc(int node)
+{
+    // Tell U-Boot to bind this node early
+    fdt_setprop_empty(dt, node, "u-boot,dm-pre-reloc");
+
+    // Make sure the power domains are bound early as well
+    int pds_size;
+    const fdt32_t *pds = fdt_getprop(dt, node, "power-domains", &pds_size);
+    if (pds) {
```
marcan, line 338: I think this won't work for >1 PD, since the FDT mutations will invalidate the pointer. You'll have to copy it out before iterating.

marcan (changes requested):

src/kboot.c
kettenis: Right.  The device tree I tested this on only has single PD dependencies.  Pushed a fix onto the branch.

marcan (approved):

PR #151, opened 2022-01-18
proxyclient: utils.py: add ehexdump()

dottedmag:
ehexdump is similar to chexdump, and uses CP437 control characters
and Braille to improve eyeballability of character output.

Signed-off-by: Mikhail Gusarov <dottedmag@dottedmag.net>


marcan:
Sounds useful for some use cases :)

PR #153, opened 2022-01-27
Add nix flake

kloenk:
This adds a nix flake as a second way to build it. [Nix](https://nixos.org) works on linux and macOS.

Currently it does not build on macOS, as clang there complains about the unknown directive `.type` in asm.


colemickens:
@Kloenk if you use this as part of a bigger nixos config for an m1 mac it would be great to take a look, thanks either way!

PR #154, opened 2022-01-27
[POC] rust support

Kloenk:
Create WDT functions written in rust.

~~Requires rust 1.59.0 (currently nightly)~~ Got it to build with stable rustc

(Not tested, as I'm to dependent on my m1 right now to reinstall the OS)


Kloenk:
Yeah, sounds fine. Only have to see how much I can do rebasing. I could also try to extend the rust support, and e.g. extend it to completely cover `add`.

marcan:
Don't worry too much about keeping it up to date; I'll probably use it to experiment with things and may just hack on the code myself or end up doing things a different way, and I'll fix any conflicts then. It's a great example of how to get things started :)

Kloenk:
The fat32 library is EOL. The author recommends https://github.com/rafalh/rust-fatfs, wich has optional no_std support

marcan:
Thanks! I pushed initial Rust chainloading support based on this PR (with the commits re-done) with a few changes (vendored crates changed to submodules, removed the WDT/ADT stuff since that code will remain C-based). Let there be Rust :-)

PR #163, opened 2022-02-07
m1n1.loadobjs: Add util for loading objects on target

povik:
For those of us who can't write assembly but still need to run code on
the other side.

```
  >>> from m1n1.loadobjs import *
  >>> lp = LinkedProgram(u)
  >>> lp.debug_printf("hello from the other side (%d)\n", 42)
  TTY> hello from the other side (42)
  0x1f
  >>> lp.load_obj("build/test.o")
  >>> lp.test()
  TTY> hello from loaded object
  0x19
  >>> lp.clear_objs()
```

--

I don't know if anybody else will find this useful but as it doesn't hook into anything maybe it can live in mainline.

### DIFF
--- a/proxyclient/m1n1/loadobjs.py
+++ b/proxyclient/m1n1/loadobjs.py
@@ -1,0 +1,163 @@
+# SPDX-License-Identifier: MIT
+from contextlib import contextmanager, ExitStack
+import sys, pathlib, os
+import subprocess
+import tempfile
+import bisect
+
+from .asm import NM, LD, OBJCOPY, ARMAsm
+
+__all__ = ["LinkedProgram"]
+
+
+def _tool_cmdline(progname, args):
+    cmdline = progname.replace("%ARCH", ARMAsm.ARCH) + " " + args
+    return cmdline
+
+def tool_output_lines(progname, args):
+    cmdline = _tool_cmdline(progname, args)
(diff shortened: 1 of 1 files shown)


povik:
Those pesky sign-offs.

svenpeter42:
What a lovely hack! :D

marcan:
Nice! I'd been meaning to do something like this for ages but it's always been in my back burner. Thanks, I'll definitely be using it :)

marcan:
If you want to add something else later: it would be cool to have a wrapper to compile a C snippet with the right include paths and such, to have it all self contained in a python script.

Also we might want to add a proxy call to get the m1n1 version and have this complain loudly if it mismatches the built one.

povik:
> On 8. 2. 2022, at 6:17, Hector Martin ***@***.***> wrote:
> If you want to add something else later: it would be cool to have a wrapper to compile a C snippet with the right include paths and such, to have it all self contained in a python script.

That’s inevitable. Seems the reasonable way to go about compiling the snippets is to still have ‘make' invoke the C compiler through some extra rule, so that the invocation matches base m1n1.

> Also we might want to add a proxy call to get the m1n1 version and have this complain loudly if it mismatches the built one.

FWIW it now compares the .rela table.

PR #167, opened 2022-03-07
experiments/dcp.py: t600x support

jannau:
Works currently via vUART under HV with dcp tracing. On bare metal dcp crashes immediately after rtkit start.This matches the behavior seen with the Linux driver although they fail with different messages in crashlog.
bare metal dcp.py: `Message 1: MMU out of tables: 200,200,0`
Linux: `Message (id=1): IOMFB int_handler_gated: failure: axi_wr_err [0x80000400]`


jannau:
The bare metal crash happens only if m1n1 was chainloaded. It works without issues when the same m1n1 binary is loaded by iboot.
The crash when started by Linux is a different problem.

PR #183, opened 2022-03-27
display: Add wait option to handle monitor disconnection

VinDuv:
Add an explicit wait for monitor reset in `display.c`, triggered by the `display=wait` (or `display=wait,<mode…>`) boot parameter.

TODO:
- ~~Check that it works on monitors other than mine~~
- ~~Check that it works with payloads~~
- Maybe improve parameter parsing so that the order of parameters doesn’t matter 
- ~~Currently the wait for the monitor to disconnect/reconnect is done between `dcp_ib_swap_begin` and `dcp_ib_swap_set_layer`.~~ The wait is now done before the swap, relying on stage1 m1n1 to wake up the monitor. However, this will fail to initialize the monitor if too much time passes between stage1 m1n1 and stage2 m1n1, allowing the monitor to fully go back to sleep between the two. That is probably not a huge concern.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -11,6 +11,7 @@
 
 #define DISPLAY_STATUS_DELAY   100
 #define DISPLAY_STATUS_RETRIES 20
+#define DISPLAY_WAIT_DELAY     10
 
 #define COMPARE(a, b)                                                                              \
     if ((a) > (b)) {                                                                               \
@@ -97,11 +98,37 @@ static int display_start_dcp(void)
     return 0;
 }
 
-int display_parse_mode(const char *config, dcp_timing_mode_t *mode)
+int display_parse_mode(const char *config, dcp_timing_mode_t *mode, int *wait_delay)
 {
     memset(mode, 0, sizeof(*mode));
+    *wait_delay = 0;
 
-    if (!config || !strcmp(config, "auto"))
+    if (!config)
+        return 0;
+
+    if (!strncmp(config, "wait", sizeof("wait") - 1)) {
+        int delay = 0;
+        config += sizeof("wait") - 1;
+        if (*config == ':') {
+            config += 1;
+            for (char c = *config; '0' <= c && c <= '9'; c = *(++config)) {
+                delay = (delay * 10) + (c - '0');
+            }
+        }
+
+        if (delay <= 0) {
+            delay = DISPLAY_WAIT_DELAY;
+        }
+
+        printf("display: wait enabled (max delay %ds)\n", delay);
+        *wait_delay = delay;
+
+        if (*config == ',') {
+            config += 1;
+        }
+    }
+
+    if (!*config || !strcmp(config, "auto"))
         return 0;
 
     const char *s_w = config;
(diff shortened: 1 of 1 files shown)


VinDuv:
I modified the patch so it only waits for disconnection before setting the mode and performing the swap, relying on the fact that the monitor is already waking up from stage1 m1n1. This works and allow the monitor to display something during the wait when a chainload is performed.

I tried to add u-boot payloads but for some reason that caused the monitor initialization to fail. Will investigate this evening.

marcan:
Gzip your U-Boot image. This is explained in the m1n1 user guide doc. It's an issue with the raw kernel format not having an accurate length field.

On 2022年3月28日 15:18:37 JST, VinDuv ***@***.***> wrote:
>I modified the patch so it only waits for disconnection before setting the mode and performing the swap, relying on the fact that the monitor is already waking up from stage1 m1n1. This works and allow the monitor to display something during the wait when a chainload is performed.
>
>I tried to add u-boot payloads but for some reason that caused the monitor initialization to fail. Will investigate this evening.
>

-- 
Hector Martin "marcan" ***@***.***)
Public key: https://mrcn.st/pub

VinDuv:
Thanks for the heads-up! That was indeed the problem. It seems to work fine with payloads now.
I modified the patch to add a configurable wait delay.
Now to see if it works with other monitors…

possiblemeatball:
Hello, good morning from EDT! Wanted to chime in with my personal experience with this patch. 

I noticed a similar display bug you are describing, with a **Samsung C27JG5X** monitor. asahi-installer would successfully complete, but no display upon reboot. This is the monitor I will be describing in this comment!

To install this patch, I have cloned the asahi-installer and edited the source to add the m1n1 var

	# (osinstall.py, asahi-installer)
	m1n1_vars.append(f"display=wait,2560x1440")

I will comment that, the base asahi-installer does not make this a rather easy process. I do not have much background with this project, admittedly--the recent Alpha release is what finally caught my eye, however I notice the installer does not allow the end-user (or expert-user for that matter) to modify m1n1 boot arguments. At minimum this patch would be more easily accessible with an accompanying asahi-installer patch, allowing the end (or expert!) user to add boot variables.

After rebooting my Mac mini with this patch applied, I can now see Arch Linux ARM with linux-asahi at full resolution, with the entire Asahi Linux Desktop package installed and functioning as well as it would otherwise.

[Wonderful experience overall](https://i.imgur.com/tYjZmk2.png), & i am hoping this will be upstreamed until GPU support makes it in!

VinDuv:
Okay, it looks like it works. The parser still requires “wait[:x],” to come before the specified resolution (if any), but I don’t think it’s worth making it more complicated to handle this.

src/display.c
```diff
@@ -97,11 +98,37 @@ static int display_start_dcp(void)
...
 }
 
-int display_parse_mode(const char *config, dcp_timing_mode_t *mode)
+int display_parse_mode(const char *config, dcp_timing_mode_t *mode, int *wait_delay)
 {
     memset(mode, 0, sizeof(*mode));
+    *wait_delay = 0;
 
-    if (!config || !strcmp(config, "auto"))
+    if (!config)
```
jannau, line 106: I would keep the mode parsing code as is. I think it makes sense to keep it as first parameters for the `display=` variable. Have a look how I added support for comma separated parameters in [#212](https://github.com/AsahiLinux/m1n1/pull/212/commits/07e54b52e6cf228fca02f973728722d241ae09c1) for `retina`

jannau, line 106: I would keep the mode parsing code as is. I think it makes sense to keep it as first parameters for the `display=` variable. Have a look how I added support for comma separated parameters in [#212](https://github.com/AsahiLinux/m1n1/pull/212/commits/07e54b52e6cf228fca02f973728722d241ae09c1)

src/display.c
```diff
@@ -130,11 +157,53 @@ int display_parse_mode(const char *config, dcp_timing_mode_t *mode)
...
+    return hpd;
+}
+
+int display_wait_disconnected(dcp_iboot_if_t *iboot, int wait_delay)
+{
+    int hpd, timing_cnt, color_cnt;
+    int max_retries = wait_delay * 1000 / DISPLAY_STATUS_DELAY;
+
+    for (int retries = 0; retries < max_retries; retries += 1) {
+        hpd = dcp_ib_get_hpd(iboot, &timing_cnt, &color_cnt);
```
jannau, line 184: `dcp_ib_get_hpd` can be called with `NULL` for timing_cnt and color_cnt, would be simpler here since we are just waiting for `hpd == 0`

jannau, line 184: `dcp_ib_get_hpd` can be called with `NULL` for timing_cnt and color_cnt

src/display.c
VinDuv: Okay, will do

marcan:
I think this is obsolete with the DCP shutdown hack.

PR #191, opened 2022-04-07
proxyclient: Add initial ProRes experiment

rqou:
This is able to use the ProRes encoder to encode YCbCr data. It depends on the t8110 DART support.


rqou:
no, that is the "AVD" block which is completely different

marcan:
I merged DART8110, can you rebase? :)

rqou:
rebased

PR #192, opened 2022-04-10
WIP: Bluetooth tracing

rqou:
Known issues:
* BAR locations and DART stream ID are hardcoded
* not sure what happens when rings wrap around
* opening the bluetooth control panel crashes for some reason


marcan:
Heh, looks like you ran into some of the same missing instructions @asahilina ran into... can you rebase the HV stuff and see if any of your changes are still needed there? (FWIW for them to be correct you'd have had to expand the Python side too, but that's done now).

rqou:
rebased, no HV changes seem to be required anymore

proxyclient/m1n1/proxy.py
```diff
@@ -712,7 +712,8 @@ def reboot(self):
     def write64(self, addr, data):
         '''write 8 byte value to given address'''
         if addr & 7:
-            raise AlignmentError()
+            self.write32(addr, data & 0xFFFFFFFF)
+            self.write32(addr + 4, data >> 32)
```
marcan, line 716: What codepath is this used in? I'd prefer to keep the alignment fault '(at least by default) since this is usually a mistake when accessing registers, though I know Apple loves using unaligned 64-bit fields in some protocols. Can we make it a separate wrapper in ProxyUtils, a subclass, or something like that?

PR #194, opened 2022-04-15
m1n1.hv: Introduce gdbserver

akihikodaki:


akihikodaki:
I have done changes I have planned for so here is a summary of changes since the initial version:

# General m1n1.hv changes
- Shadow used debug registers, fixing debugging Linux.
- Allow to switch CPU without continuing.
- Always continue on the stepped thread, aligning with GDB's semantics (otherwise GDB crashes).
- Propagate breakpoint change to other CPUs.
- Add hardware watchpoint support.

# gdbserver-specific changes
- Interrupt a CPU when connected with GDB, otherwise GDB connection times out.
- Support switching threads.
- Support watchpoints.
- Add missing `SPDX-License-Identifier: MIT`.
- Minor code changes.

marcan:
Thank you for working on this! This is a really cool feature I wanted to have for a long time.

Can you fix the conflicts on hv.py, and maybe explain a bit how the exception handling / CPU switching stuff changes? From what I can tell:

* The GDB stub runs in its own thread
* The HV has its existing shell stuff largely as-is
* The SIGUSR1 hack is introduced to allow the `cont()` / `cpu()` stuff to cross thread boundaries from GDB to the main thread
* The in-thread CPU switching relies on this thread split so the HV thread can exit back out to the proxy path for CPU switching, while the GDB thread continues in line

I've long been bothered by the cpu() stuff causing an exit out of the shell. If I figure out a better way to implement it without an unwind (e.g. with proxy changes so the context switch happens without an exit out to the main loop, at least from the Python perspective) would that help make this a bit simpler?

marcan:
I just pushed some refactoring that makes `cpu()` not exit the shell context, and the same approach can hopefully be extended to do things like single-step if needed. Do you think you can use this to make your life easier, and maybe avoid the SIGUSR1 hack? :)

marcan:
Though I guess you'd still need a signal or something similar to exit the shell if you want synchronize the gdbserver and hv proxy shell execution state and allow using both at once (which I guess might be handy?) while still supporting break/continue from both sides. But at least it should simplify the CPU switching stuff.

Alternatively you could have the gdbserver stuff mutex with the shell, so only one side can "own" the HV (when paused) at once.

src/hv_exc.c
```diff
@@ -38,6 +38,7 @@ static u64 exc_entry_time;
 
 extern u32 hv_cpus_in_guest;
 extern int hv_want_cpu;
+extern bool hv_want_cpu_in_rendezvous;
```
marcan, line 41: Instead of doing it like this, how about we add a separate int called `hv_pinned_cpu`?

src/hv_exc.c
```diff
@@ -91,12 +92,13 @@ static void _hv_exc_proxy(struct exc_info *ctx, uartproxy_boot_reason_t reason,
     }
 }
 
-static void hv_maybe_switch_cpu(struct exc_info *ctx, uartproxy_boot_reason_t reason, u32 type,
```
marcan, line 94: Let's keep the new name here (after my refactoring); rendezvous isn't really what this does.

src/hv_exc.c
```diff
@@ -91,12 +92,13 @@ static void _hv_exc_proxy(struct exc_info *ctx, uartproxy_boot_reason_t reason,
...
-                                void *extra)
+static void hv_rendezvous_cpu(struct exc_info *ctx, uartproxy_boot_reason_t reason, u32 type,
+                              void *extra)
 {
-    while (hv_want_cpu != -1) {
+    while (hv_want_cpu_in_rendezvous && hv_want_cpu != -1) {
```
marcan, line 98: If there's a separate pinned CPU flag, then this function doesn't have to change at all.

src/hv_exc.c
```diff
@@ -111,10 +113,25 @@ static void hv_maybe_switch_cpu(struct exc_info *ctx, uartproxy_boot_reason_t re
 void hv_exc_proxy(struct exc_info *ctx, uartproxy_boot_reason_t reason, u32 type, void *extra)
 {
     /*
-     * If we end up in the proxy due to an event while the host is trying to switch CPUs,
-     * handle it as a CPU switch first. We still tell the host the real reason code, though.
+     * Wait until other CPUs finish processing requests or this CPU gets requested.
+     * If a CPU switch in rendezvous is requested, handle it before actually handling the
+     * exception. We still tell the host the real reason code, though.
      */
-    hv_maybe_switch_cpu(ctx, reason, type, extra);
```
marcan, line 117: We can keep this here, because it won't do anything if `hv_want_cpu` is false...

src/hv_exc.c
```diff
@@ -424,7 +441,7 @@ void hv_exc_fiq(struct exc_info *ctx)
         tick = true;
     }
 
-    if (mrs(TPIDR_EL2) != 0 && !(mrs(ISR_EL1) & 0x40) && hv_want_cpu == -1) {
```
marcan, line 427: Then this needs to check for `hv_want_cpu == -1 && hv_pinned_cpu != mrs(TPIDR_EL2)`, and the below check for `mrs(TPIDR_EL2) == 0` also needs to change to `== hv_pinned_cpu` only if it is not -1.

marcan (changes requested):

marcan:
Thanks! Looks good to me, if you can fix the merge conflicts I'll merge it! I think it's just an artifact of having moved `hv.py`, since what Lina pushed yesterday shouldn't clash changes linewise.

marcan (approved):

akihikodaki:
e1a1419ea6e004b1a853172977029651f9864c55: rebased to 7ed86d89bb4e7674206514611fb8d6a5a7037089.

akihikodaki:
c6e60c238be35e18de7bb7a0d180d738ac7e6db8: Added initialization of `hv_pinned_cpu` to `hv_start` and converted argument -1 for `hv_pin_cpu` to `0xffffffffffffffff` since `M1N1Proxy` encodes arguments as unsigned.

I forgot to recompile m1n1 so the code has actually not been tested at all(!). Fortunately the two changes were enough to get it work. (Background: I use a Makefile which automatically compiles the binary before launching m1n1 for Linux kernel debugging, but I don't use it for Darwin kernel debugging.)

marcan:
Thanks again! FWIW my approach with the hypervisor is to move fast and break things, so I'm not worried if there are any regressions lurking, we'll fix them in time. Sorry for taking so long to review this - it was large enough that I wanted to take a decent look and I've been busy, but I'm glad it's in!

PR #199, opened 2022-04-19
thunderbolt tunables + change linux format to be forwards-compatible

svenpeter42:


danieltarazona:
@asahilina @svenpeter42  @bastilian Could you prioritize Thunderbolt support too? Now that GPU driver is ready I would like to see eGPU's support return again

svenpeter42:
I’m not gonna prioritize anything and eGPUs are unlikely to work even with thunderbolt support.

brianathere:
Historical pointer for future readers: the `tbt-tunables` work here (and its successor PR #275 that merged the ACIO tunable + DROM propagation) landed the m1n1-side groundwork that would have been needed for a Linux ACIO driver, but the Linux-side driver has never existed because the ACIO host-mode bring-up path appears to be gated at the SoC fabric level on M1 Pro/Max and does not appear reachable from research-tier callers based on the testing I did. I wrote up the full wall — the exact `IOMappedWrite32` that macOS's `enableEmbeddedCPU` issues, the empirical confirmation that NVRAM + `bputil` + Permissive Security + AMFI bypass don't lift it, and the speculation about SPTM's role — at AsahiLinux/linux#479. This comment exists only so somebody who comes back to this PR in three years knows to read that issue before spending another week RE'ing the register map.
— brianathere

marcan:
(The above comment is an LLM bot. Thunderbolt works.)

PR #211, opened 2022-05-27
rtkit/dcp: Ensure the AP -> IOP channel is ready for shutdown

jannau:
Pending syslog messages delivered during shutdown can not be
acked since shutdown command is still in the mailbox. The missing ack
blocks the mailbox communication.
Draining any pending messages before shutting down fixes this.

The issue is reproducible on the Mac Studio with a Dell UP2414Q display.
If the display is in power save mode when the mode is set it resets the
link ~400ms. The reset produces syslog messages which are not read
before shutdown.

### DIFF
--- a/src/afk.c
+++ b/src/afk.c
@@ -459,8 +459,18 @@ err:
     return NULL;
 }
 
+#define ASC_TIMEOUT_US 50
+
 int afk_epic_shutdown(afk_epic_ep_t *epic)
 {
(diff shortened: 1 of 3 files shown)


marcan:
I'm confused, how does the pending shutdown block the syslog acks? The mailbox is a FIFO, it should have plenty of space for the syslog ACKs to go in after the shutdown request.

Either way, asc_send() definitely needs a timeout. Something like 1 second is sufficient; we should really never hit the case where that blocks unless something went horribly wrong.

PR #213, opened 2022-05-31
experiments/ohmmeter.py: Make CS42L84 your ohmmeter

povik:
Add script which prints out the (rough) measured impedance of connected
headphones on machines with the CS42L84 jack codec (2021 Macbook Pros).

    $ proxyclient/experiments/ohmmeter.py
    Waiting for plug... measuring... 64.3 ohms... yanked
    Waiting for plug... measuring... 697.1 ohms...

Signed-off-by: Martin Povišer <povik@protonmail.com>

### DIFF
--- a/proxyclient/experiments/ohmmeter.py
+++ b/proxyclient/experiments/ohmmeter.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: MIT
+import sys, pathlib
+import time
+sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
+
+from m1n1.utils import *
+from enum import IntEnum
+from m1n1.setup import *
+from m1n1.hw.i2c import I2C
+
+
+class R_IRQ_MASK1(Register8):
+    RING_PLUG   = 0
+    RING_UNPLUG = 1
+    TIP_PLUG    = 2
+    TIP_UNPLUG  = 3
+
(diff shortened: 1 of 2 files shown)


povik:
Fixed sign-offs (email confusion...)

proxyclient/experiments/ohmmeter.py
```diff
@@ -0,0 +1,260 @@
...
+l84 = CS42L84(i2c2, 0x4b)
+
+init_ring_tip_sense()
+
+while True:
+    print("Waiting for plug... ", end=""); sys.stdout.flush()
+    wait_for_plug()
+
+    print("measuring... ", end=""); sys.stdout.flush()
+    print(f"{sense_Z():.1f} ohms... ", end=""); sys.stdout.flush()
```
mcpower, line 257: You can use the [`flush` keyword argument of `print`](https://docs.python.org/3/library/functions.html#print) to flush stdout instead of calling `sys.stdout.flush()`:

```suggestion
    print("Waiting for plug... ", end="", flush=True)
    wait_for_plug()

    print("measuring... ", end="", flush=True)
    print(f"{sense_Z():.1f} ohms... ", end="", flush=True)
```

povik: Ha, forgot `print` can do that, thanks! (Will not be doing a fix right away since `experiments/` don't need to be particularly high quality.)

PR #217, opened 2022-06-06
Display configuration: Multiple fixes

VinDuv:
* Improve parsing of display options so they work without specifying a resolution (i. e. `display=retina` works).
 * Fix handling of framebuffer reset when switching between non-retina and retina modes without changing resolution.
 * Change display init/config code to only initialize external displays, and only configure them once.


marcan:
Closing since this is really old and a lot of display changes happened since then. If this is still relevant please rebase and reopen another PR.

PR #219, opened 2022-06-10
hv/trace_mesa.py: add mesa tracer

chadmed:
Probably worth treating this more like an RFC than a proper PR,
since this doesn't actually do anything particularly useful without the
SEP tracer. 

The only really useful stuff you can get out of this is:

* A pointer to an address which _seems_ to store the actual fingerprint image
* The fingerprint image (though I haven't tried undumping this into a binary file)
* That the command buffer is encrypted
* Its behaviour is consistent across reboots, though the data found in its memory isn't
* There is a SIO endpoint (0x2) that seems to send/receive data when a finger is placed on the
   sensor, but trying to read from it in the same way as other EPs crashes the hypervisor

We could probably systematise the reading of the IOVA regions, but I've just been running
iomon.poll() after halting the guest with my finger on the sensor to see what changes.=

### DIFF
--- a/proxyclient/hv/trace_mesa.py
+++ b/proxyclient/hv/trace_mesa.py
@@ -1,0 +1,218 @@
+# SPDX-License-Identifier: MIT
+"""
+Things to note:
+    The command buffer is encrypted after the poweron sequence, and I
+    can't find the key in the SEP using sven's old SEP tracer.
+"""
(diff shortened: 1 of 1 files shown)


marcan:
Works for me, no need to be picky about HV experiments :)

PR #222, opened 2022-06-18
utils: Add 'sev' in spin_unlock()

jannau:
Using the global monitor to generate an event on unlock of a spinlock
does not work with the MMU disabled. This fixes a regression from
9c795fbdbf445d144d331ff2a19a0f42fe0fc190 and an incomplete fix from
816388325262e5c7dd681c8ac6392248b6ca9f0e.

Bisected and fix confirmed by Sven.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -168,8 +168,11 @@ void spin_unlock(spinlock_t *lock)
     s64 me = smp_id();
     assert(__atomic_load_n(&lock->lock, __ATOMIC_RELAXED) == me);
     assert(lock->count > 0);
-    if (!--lock->count)
+    if (!--lock->count) {
         __atomic_store_n(&lock->lock, -1L, __ATOMIC_RELEASE);
+        // the store doesn't trigger an event from the global monitor with the MMU disabled
+        sysop("sev");
+    }
 }
 
 bool is_heap(void *addr)


jannau:
@akihikodaki please check if this change regresses your use cases. @svenpeter42 and I see without this change a very slow guest execution from the point of m1n1 disables the MMU until some point in the payload (Linux) execution.

svenpeter42 (approved):

akihikodaki:
@jannau If SEV is necessary anyway, using exclusive lock makes no sense. In theory, it should be simply reverted to the state of commit 9c795fbdbf445d144d331ff2a19a0f42fe0fc190 with a DSB instruction added as described in commit 816388325262e5c7dd681c8ac6392248b6ca9f0e.

However, that did not work for me in reality as described in commit 816388325262e5c7dd681c8ac6392248b6ca9f0e. On my MacBook Air, WFE missed SEV and waited too long and failed to launch cores when booting HV even with a DSB instruction added. Can you reproduce the problem? Unfortunately I couldn't figure out what is going on so it would be nice if you can reproduce and debug it.

marcan:
Spinlocks have never worked with the MMU disabled, that's why the iodev code elides them in that case. Under what circumstances are we taking locks with the MMU disabled here? If we are, that needs to be fixed.

svenpeter42:
Looks like the slowdown are actually caused smp_set_wfe_mode here: If I have no sev in the unlock and the guest m1n1 calls smp_set_wfe_mode everything slows down to a crawl here.  If I remove that call however everything remains fast.
I'll see if I can actually figure out what causes this.

jannau:
cause is not in the spinlock so this change merely made the symptom disappear

marcan:
Okay, that makes more sense. The HV is probably interacting badly with the events from the guest. The weird thing is though, the ARM ARM says the event register is set by an exception return so... what gives?

marcan:
Is this a data ordering bug in `smp_set_wfe_mode` by any chance? Does it go away if you change the `dmb sy` there to a `dsb sy` and/or an `isb`?

marcan:
I just pushed a commit that may or may not help things (specifically needed for the guest m1n1), can someone test and see if it helped?

jannau:
The commit (and additional `isb`/`dsb sy` in smp.c) doesn't fix the slowdown.

The issue seems to be that HV m1n1 doesn't get out of the `wfe` in `spin_lock` when it should (removing the `wfe` avoids the issue). Judging by the time in the kernel log we spend the time in the hypervisor (time doesn't progress although it is incredibly slow).

marcan:
Next idea: s/`ldxr`/`ldaxr`/ in spin_lock.

marcan:
Basically my theory is the CPU is speculating through two loops of `ldxr/wfe` since `ldxr` doesn't have acquire semantics, so it doesn't introduce ordering.

```
ldxr (lock is locked)
    exclusive monitor set
ldxr (speculative, lock is locked)
    exclusive monitor remains set
(lock becomes unlocked here)
    exclusive monitor becomes clear, event set
wfe
    event is cleared
wfe
    stuck```

akihikodaki:
@marcan I don't think that happens since the loop contains `casa`. I have not confirmed that though (I broke my x86 PC so I cannot run proxyclient :/)

akihikodaki:
Ah, but `casa` may not enforce the acquire semantics if it failed.

jannau:
no change with `ldaxr` and even an `isb` after `wfe` makes no difference

akihikodaki:
I rechecked the reference manual and it seems `casa` always enforce the acquire semantics.

marcan:
> The issue seems to be that HV m1n1 doesn't get out of the `wfe` in `spin_lock` when it should (removing the `wfe` avoids the issue).

This might be a red herring. What happens if you do a "sevl" at the end of spin_lock, leaving the wfe?

jannau:
no, `sevl` instead of `sev` in spin_unlock() is still slow

marcan:
Fixed. It had nothing to do with the spinlock. The SMP WFE codepath was just pathologically thrashing the hypervisor.

marcan:
As for why this all had to do with the spinlock change: it probably changed the fairness characteristics of the lock, where previously CPU#0 had better chances and things didn't slow down to a crawl since the guest could make forward progress at a reasonable pace until it initializes SMP and takes the secondaries out of the WFE loop. After the change CPU#0 got the short end of the stick more often and the problem, which was always there, became evident as a massive slowdown.

PR #225, opened 2022-06-22
add raw binary support to HV scripts

amarioguy:
Hello!

Had to repost because the last pull request had issues, also cleaned up some code

As part of my efforts to boot Windows on Apple silicon Macs successfully, part of that meant adding raw binary support to the hypervisor scripts (This is to allow raw binary UEFI images to be executed directly using the run_guest.py script instead of being forced to wrap them in Mach-O headers or concatenate them to a Mach-O wrapped m1n1 payload, allowing for quicker build and test cycles)

I wanted to contribute this change back to mainline m1n1 as I feel that it'd be a really nice to have addition to the hypervisor so that everyone can benefit from being able to directly load raw binaries using the run_guest script.

Most of the code from the load_macho logic ended up working perfectly for the load_raw logic, albeit with some minor changes to account for lack of a Mach-O header and whatnot, so this was a fairly minor change albeit one that I felt would be really beneficial both in my work and in the community's hopefully. I did move out all the common code between Mach-O and raw binary code into a load_common method and have both load_macho and load_raw call that, this ended up working perfectly and helped reduce redundant code

Tested on an m1n1 payload, both raw and Mach-O wrapped, as well as a macOS kernelcache.

This is my first "official" contribution to the project, hopefully it helps the project!

Signed-off-by: amarioguy <arminders208@outlook.com>

### DIFF
--- a/proxyclient/m1n1/hv/__init__.py
+++ b/proxyclient/m1n1/hv/__init__.py
@@ -1487,9 +1487,8 @@ class HV(Reloadable):
     def disable_time_stealing(self):
         self.p.hv_set_time_stealing(False)
 
-    #only meant to be called from load_raw or load_macho
-    #macho_data is unused for raw images, but not sure if there's a more "elegant" solution
-    def load_common(self, image, is_macho_bin, macho_data):
+
+    def load_raw(self, image, entryoffset=0x800):
         sepfw_start, sepfw_length = self.u.adt["chosen"]["memory-map"].SEPFW
         tc_start, tc_size = self.u.adt["chosen"]["memory-map"].TrustCache
         if hasattr(self.u.adt["chosen"]["memory-map"], "preoslog"):
@@ -1522,10 +1521,8 @@ class HV(Reloadable):
 
         print(f"Physical memory: 0x{phys_base:x} .. 0x{mem_top:x}")
         print(f"Guest region start: 0x{guest_base:x}")
-        if(is_macho_bin == True):
-            self.entry = macho_data.entry - macho_data.vmin + guest_base
-        else:
-            self.entry = guest_base + 0x800
+        
+        self.entry = guest_base + entryoffset
 
         print(f"Mapping guest physical memory...")
         self.add_tracer(irange(self.ram_base, self.u.ba.phys_base - self.ram_base), "RAM-LOW", TraceMode.OFF)
@@ -1571,9 +1568,8 @@ class HV(Reloadable):
             addr, size = cpu.cpu_impl_reg
             print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
             self.p.write64(addr, rvbar)
-    def load_raw(self, data):
-        image = data
-        self.load_common(image, False, 0)
+
+
     def load_macho(self, data, symfile=None):
         if isinstance(data, str):
             data = open(data, "rb")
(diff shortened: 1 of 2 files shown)


proxyclient/m1n1/hv/__init__.py
```diff
@@ -1564,8 +1522,10 @@ def load_hook(data, segname, size, fileoff, dest):
 
         print(f"Physical memory: 0x{phys_base:x} .. 0x{mem_top:x}")
         print(f"Guest region start: 0x{guest_base:x}")
-
-        self.entry = macho.entry - macho.vmin + guest_base
+        if(is_macho_bin == True):
```
marcan, line 1525: Style: no parentheses around the condition in Python

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1487,51 +1487,9 @@ def enable_time_stealing(self):
...
-                    print(f"  0x{off:x}: 0x{opcode:04x} -> hvc 0x{opcode:x} (0x{inst:x})")
-                    a[p // 4] = inst
-                p += 4
-
-            print("Done.")
-            return a.tobytes()
-
-        #image = macho.prepare_image(load_hook)
-        image = macho.prepare_image()
+    #only meant to be called from load_raw or load_macho
```
marcan, line 1490: Style: please leave a blank line between methods (I know we're not 100% consistent with this, but especially for complex classes like this one it helps).

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1564,8 +1522,10 @@ def load_hook(data, segname, size, fileoff, dest):
 
         print(f"Physical memory: 0x{phys_base:x} .. 0x{mem_top:x}")
         print(f"Guest region start: 0x{guest_base:x}")
-
-        self.entry = macho.entry - macho.vmin + guest_base
+        if(is_macho_bin == True):
+            self.entry = macho_data.entry - macho_data.vmin + guest_base
```
marcan, line 1526: Why not just pass an `entry` parameter instead of the macho_data thing? That would be relative to the guest_base. Then here you just need to add guest_base, and you don't need any specific logic for macho/not.

At that point, this entire function can just become `def load_raw(self, data, entry=0x800):` itself, instead of `load_common`. Then `load_macho` can just call `load_raw` directly.

marcan (changes requested):

amarioguy:
Alright, I've pushed a very slight refactor of the code, so that it meets the style/readability guidelines, and yes, load_common is now load_raw with an extra parameter for the entry point offset (for clarity: entryoffset is named as such to help distinguish between it and self.entry in the actual method)

proxyclient/m1n1/hv/__init__.py
amarioguy: resolved with the refactor

proxyclient/m1n1/hv/__init__.py
amarioguy: resolved with the refactor

proxyclient/m1n1/hv/__init__.py
amarioguy: resolved with the refactor

marcan:
Thanks!

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1603,8 +1560,6 @@ def load_hook(data, segname, size, fileoff, dest):
         self.tba.devtree = self.adt_base - phys_base + self.tba.virt_base
         self.tba.top_of_kernel_data = guest_base + image_size
 
-        self.sym_offset = macho.vmin - guest_base + self.tba.phys_base - self.tba.virt_base
```
svenpeter42, line 1606: @marcan @amarioguy   This commit breaks symbols for XNU kernels, probably because sym_offset is always zero now. I've revered it locally for now.

PR #228, opened 2022-06-29
hv: Add 9P virtio peripheral

povik:
This allows one to access host m1n1 sources from within the system running in hypervisor. Once you arrange for the virtio peripheral both in hypervisor code (`self.attach_virtio(Virtio9PTransport(), 0x2_2000_0000, 382)`) and in the devicetree, you run

    mount -t 9p -o transp=virtio m1n1 /mnt

to mount the sources on guest. One use case of it is using m1n1 regmaps over kernel's internal regmap structures (as alluded to by the extra commit).

Performance is not great since the hypervisor stops to handle each 9P call but it's not impractical.

Direction I think this should be expanded in:

 * Allocate the IRQs/register range automatically
 * Insert the peripheral into ADT
 * Have guest m1n1 transfer it from ADT to FDT

Let me know if you think this is fine for inclusion.

### DIFF
--- a/src/kboot.c
+++ b/src/kboot.c
@@ -1524,6 +1524,74 @@ err:
     return ret;
 }
 
+static int dt_transfer_virtios(void)
+{
+    int path[3];
+    path[0] = adt_path_offset(adt, "/arm-io/");
+    if (path[0] < 0)
+        bail("ADT: /arm-io not found\n");
+
+    int aic = fdt_node_offset_by_compatible(dt, -1, "apple,aic");
+    if (aic == -FDT_ERR_NOTFOUND)
+        aic = fdt_node_offset_by_compatible(dt, -1, "apple,aic2");
+    if (aic < 0)
+        bail("FDT: failed to find AIC node\n");
+
+    u32 aic_phandle = fdt_get_phandle(dt, aic);
+    const fdt32_t *ic_prop = fdt_getprop(dt, aic, "#interrupt-cells", NULL);
+    u32 intcells = 0;
+    if (ic_prop)
+        intcells = fdt32_ld(ic_prop);
+    if (intcells < 3 || intcells > 4)
+        bail("FDT: bad '#interrupt-cells' on AIC node (%d)\n", intcells);
+
+    for (u32 i = 0; i < 16; i++) {
+        char name[16], fullname[32];
+        snprintf(name, sizeof(name) - 1, "virtio%d", i);
+
+        path[1] = adt_subnode_offset(adt, path[0], name);
+        if (path[1] < 0)
+            break;
+        path[2] = 0;
+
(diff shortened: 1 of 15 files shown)


marcan:
Very cute. I like it :)

I was going to say I'm not sure about putting this in the ADT, but on further thought there's no reason not to do that (as long as the mutation happens in Python, since we already have code for that there). So sure, why not!

marcan:
Just sort of direction-wise: I'm fine with anything in m1n1/the hypervisor as long as it helps with development. I don't think it should become a general-purpose hypervisor or veer off towards production usage, but sharing files with the host is *very* obviously useful for a ton of reasons (e.g. you can avoid building initramfses all the time for development, develop kernel modules and load/reload them from the guest without a full reboot, etc.). I've been doing this over sshfs in the past (host to guest, which sucks a bit), but the other way around with 9p should be nicer for a bunch of workflows.

povik:
Okay, I implemented the hypervisor->ADT->FDT path to pass information on the existence of the virtio device, and added allocation of IRQ/register range. The allocation looks for numbers/ranges that are free per ADT -- for MMIO that should be foolproof, and I guess the AIC isn't to any degree shared with the coprocessors, so for IRQs that should be okay too.

So, about usage: One passes a path and a tag to the hypervisor, so for example

    ./proxyclient/tools/run_guest.py \
        -v ~/repos/m1n1:m1n1 \
        -v ~/music:tunes \
        m1n1-payload.macho

to export one tree under tag `m1n1` and another under `tunes`. To then mount the tags on guest:

    mount -t 9p -o transp=virtio tunes /mnt
    mount -t 9p -o transp=virtio m1n1 /mnt2

You need a `u9fs` program installed on the host to do the 9P speaking, and these options on the guest kernel (let me know anybody if some are missing):

    CONFIG_NET_9P
    CONFIG_NET_9P_VIRTIO
    CONFIG_VIRTIO
    CONFIG_VIRTIO_MMIO
    CONFIG_9P_FS

src/kboot.c
```diff
@@ -1114,6 +1182,10 @@ int kboot_prepare_dt(void *fdt)
         return -1;
     if (dt_disable_missing_devs("i2c", "i2c@", 8))
         return -1;
+#ifndef RELEASE
```
povik, line 1185: Not sure if we want !RELEASE or some other gate here

povik:
Rebased

povik:
Rebased

povik:
Ping @marcan

marcan:
Looks good, can you fix the conflict so I can merge?

povik:
Sure, thanks for the review!

povik:
Rebased (and removed a stray definition in `uartproxy.h`)

PR #233, opened 2022-07-16
Initial hardware scaler experiment

rqou:
This is a first pass at an experiment to use the hardware scalers. It can resize images in BGRA/RGBA linear order.

The details of how the scaling arithmetic gets performed is not understood yet. Also, many other advanced features are not understood, such as

* other input/output formats
* histograms
* color space conversion
* chroma subsampling
* "ASE"

This experiment uses the "main" block of registers with index 0 in the ADT. There exists a "APIODMA" block that appears to be able to program scaler settings based on a command list, but macOS does not use this, and it is not yet clear whether or not this functionality actually works.


alyssarosenzweig:
Do we know what the scalers are actually intended for? when would they be used (on linux or macos) instead of scaling on the display controller or GPU? I guess maybe hardware video encoding?

alyssarosenzweig:
`XXX this is a random filter grabbed from a random trace` I hope this can be deleted :sweat_smile:

rqou:
re where the scalers are used: I am not sure, but I saw a small number of uses on boot as the desktop loaded. the hardware seems to generally back "IOSurfaceAccelerator"

re filter coefficients: this is some kind of impulse response function, but I am not sure exactly how the arithmetic is performed or what order the coefficients go in. once that is figured out, this can be replaced with a calculated sinc function or other antialiasing filter

unfortunately I will not be able to do any further testing on this for a while, since my laptop got damaged at May Contain <strike>Hardware Failure</strike>Hackers

PR #245, opened 2022-08-27
HV: add support to load additional kext symbols

svenpeter42:


proxyclient/m1n1/hv/__init__.py
```diff
@@ -1683,6 +1683,54 @@ def load_system_map(self, path):
                 self.symbol_dict[name] = addr
         self.symbols.sort()
 
+    def add_kext_symbols(self, data, kextname=None, demangle=False):
```
povik, line 1686: This is similar to what `MachO.add_symbols` does (including the relocation), only it loads symbols for all sections. I suggest we just refer to it, so that we save precious `hv/__init__.py` lines. So something like (which I have written recently for myself):

```
    def _load_macho_symbols(self):
        self.symbol_dict = self.macho.symbols
        self.symbols = [(v, k) for k, v in self.macho.symbols.items()]
        self.symbols.sort()

    def add_kext_symbols(self, kext):
        import plistlib
        info_plist = plistlib.load(open(f"{kext}/Contents/Info.plist", "rb"))
        identifier = info_plist["CFBundleIdentifier"]
        name = info_plist["CFBundleName"]
        macho = MachO(open(f"{kext}/Contents/MacOS/{name}", "rb"))
        self.macho.add_symbols(identifier, macho)
        self._load_macho_symbols()
```

PR #256, opened 2022-09-30
preserve iommu mappings for DCP/disp0

jannau:
- Add the boot iommu mappings for dcp/disp0 to the FDT based on https://lore.kernel.org/asahi/20220923123557.866972-1-thierry.reding@gmail.com
- lock dart-disp0

Matching kernel tree is https://github.com/jannau/linux/tree/asahi-dcp-september

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,6 +90,7 @@ OBJECTS := \
 	dart.o \
 	dcp.o \
 	dcp_iboot.o \
+	devicetree.o \
 	display.o \
 	exception.o exception_asm.o \
 	fb.o font.o font_retina.o \
--- a/src/dart.c
+++ b/src/dart.c
@@ -559,10 +559,10 @@ u64 dart_search(dart_dev_t *dart, void *paddr)
         }
     }
 
-    return 0;
+    return DART_PTR_ERR;
 }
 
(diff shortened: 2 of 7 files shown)


jannau:
Hej @marcan,
`/reserved-memory` nodes and the `iommu-addresses` / `memory-region` links are now gone from the linux *.dts and *.dtsi and are added by m1n1 instead. The only chand in the dts source is adding aliases for disp0, disp0_piodma, dcp and dcpext*.
disp0_piodma is a little bit annoying since the dcp driver includes just an empty device for that iommu. I don't see a nicer solution on linux side though. It would be nice if the dcp device could have 2 independent iommus.
One thing I should try is if we need the initial disp0-piodma mapping. The display iommu initialization on the mac studio ignores disp0-piodma without causing issues so it might not needed. Not sure how to verify that with certainty.

Anyway, is this going in the right direction? It would be nice to have this in m1n1 to make experimenting with dcp in the kernel easier. @asahilina will need something similar for UAT. See https://github.com/jannau/m1n1/commit/19c22f850a60e82997e6a9e7779b8abb5fa21110 for an initial implementation allowing the agx driver to work on m1 systems with 16GB RAM.

marcan:
I'm happy merging this if you're happy with it, let's get the DCP ball rolling. Feel free to drop the draft tag :)

PR #257, opened 2022-10-16
fix(proxy): fix hv translate case

ElijahPepe:
The P_HV_TRANSLATE case calls hv_translate(), which requires the fourth parameter to be a pointer to a u64. This line also fails lint tests; the fourth parameter should be on a new line.

Signed-off-by: Elijah Conners <business@elijahpepe.com>


marcan:
This fix is wrong (it should be taking the argument as a pointer), but it's also already fixed in main.

PR #259, opened 2022-10-23
Tapping the internal microphones

povik:
This has extension of `experiments/aop.py`, addition of AOP tracer and some other tweaks.

To cite the key commit:

    experiments/aop.py: Extend AOP experiment, add tracer

    The AOP uses an 'EPIC' protocol similar to the one other coprocessor
    firmware is using but not in the exact same version. Add code for
    tracing the AOP calls and extend the aop.py experiment with the client
    side of it. Include description of audio calls and some other calls
    related to sensor discovery.

    Furthermore, in experiments/aop.py, do some AOP audio setup. Once that
    is done we can start streaming samples from the internal microphones by
    making what AOP considers power state adjustment calls. That is, we
    adjust the power state of a 'hpai' device, first to a 'pw1 ' stage,
    then to 'pwrd' stage.

    So, to see microphone samples, enter the AOP experiment shell first:

      $ M1N1DEVICE=/dev/ttyACM0 experiments/aop.py

    Within the shell, adjust the power state of 'hpai':

      >>> aop_set_audio_pstate('hpai', 'pw1 ')

    At that point /arm-io/admac-aop-audio powers up. In parallel to the AOP
    shell, we can start tools/admac_stream.py on the just powered-up ADMAC
    instance:

      $ M1N1HEAP=0x10010000000 M1N1DEVICE=/dev/ttyACM1 tools/admac_stream.py \
            --node admac-aop-audio --channel 1 -v | xxd -g 4 -c 12 -e

    Returning back to the AOP shell, we can then set 'hpai' to 'pwrd' state
    to kick off the streaming:

      >>> aop_set_audio_pstate('hpai', 'pwrd')

    By that point, we should see samples coming out on the ADMAC end. The
    samples are 32-bit floats packed in groups of three in a frame, e.g.

    00000000: ba7ac6a7 ba32d3c3 baa17ae2  ..z...2..z..
    0000000c: 38ccea5f b99c1a37 ba0c4bb1  _..87....K..
    00000018: 39d2354f 3964b5ff 39b209fb  O5.9..d9...9
    00000024: b96a1d1f 39c8503f 3958fc4f  ..j.?P.9O.X9
    00000030: b6b1f5ff 39c72b8f 39bbe017  .....+.9...9
    0000003c: 3a912de5 36dd4f7f 37f1147f  .-.:.O.6...7

    This has been tested and will to some degree be specific to 2021 Macbook
    Pro (t6000). Differences on other models TBD (at the very least the
    number of microphones can be presumed different).

    Signed-off-by: Martin Povišer <povik@protonmail.com>

### DIFF
--- a/proxyclient/experiments/aop.py
+++ b/proxyclient/experiments/aop.py
@@ -4,6 +4,7 @@ import sys, pathlib
 sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
 
 import struct
+import traceback
 from construct import *
 
 from m1n1.setup import *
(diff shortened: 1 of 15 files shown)


marcan:
Ha, so AOP uses the same IOP ring buffer/EPIC endpoints with a different tag? Finally some consistency...

PR #263, opened 2022-11-12
Small improvement to tethered boot support from macOS host

aderuelle:
Small improvement to the out-of-the-box experience when using maOS as tethered boot host

- use `/dev/cu.usbmodemP_03` and `/dev/cu.usbmodemP_01` for `picocom-sec.sh` and `proxyclient` respectively

### DIFF
--- a/proxyclient/m1n1/proxy.py
+++ b/proxyclient/m1n1/proxy.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: MIT
-import os, sys, struct, serial, time
+import platform, os, sys, struct, serial, time
 from construct import *
 from enum import IntEnum, IntFlag
 from serial.tools.miniterm import Miniterm
@@ -131,13 +131,18 @@ class UartInterface(Reloadable):
     REPLY_LEN = 36
     EVENT_HDR_LEN = 8
 
+    DEFAULT_UART_DEV="/dev/m1n1"
+    DEFAULT_BAUD_RATE=115200
+    if platform.system() == 'Darwin':
+        DEFAULT_UART_DEV="/dev/cu.usbmodemP_01"
+
     def __init__(self, device=None, debug=False):
         self.debug = debug
         self.devpath = None
         if device is None:
-            device = os.environ.get("M1N1DEVICE", "/dev/m1n1:115200")
+            device = os.environ.get("M1N1DEVICE", self.DEFAULT_UART_DEV)
         if isinstance(device, str):
-            baud = 115200
+            baud = self.DEFAULT_BAUD_RATE
             if ":" in device:
                 device, baud = device.rsplit(":", 1)
                 baud = int(baud)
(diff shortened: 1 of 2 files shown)


marcan:
LGTM, feel free to unmark as draft.

martell:
can confirm this works for me out of the box on both a macos and linux host, thank you @aderuelle I carry this on my local envs

proxyclient/m1n1/proxy.py
```diff
@@ -131,13 +131,20 @@ class UartInterface(Reloadable):
     REPLY_LEN = 36
     EVENT_HDR_LEN = 8
 
+    DEFAULT_UART_DEV="/dev/m1n1"
+    DEFAULT_BAUD_RATE="115200"
```
marcan, line 135: This needs to be an int

proxyclient/tools/picocom-sec.sh
```diff
@@ -1,7 +1,30 @@
...
+#
+# Secondary UART device name exposed when m1n1 boots is /dev/m1n1-sec on
+# Linux host and /dev/cu.usbmodemP_03 on macOS host
+#
+DEFAULT_SECDEV=/dev/m1n1-sec
+PLATFORM_NAME=$(uname)
+
+case "${PLATFORM_NAME}" in
+Linux)
+    SECDEV=/dev/ttyACM1 ;;
```
marcan, line 11: Same here, the udev rules make `/dev/m1n1-sec` work on Linux (more reliably than guessing ACM device numbers).

proxyclient/m1n1/proxy.py
```diff
@@ -131,13 +131,20 @@ class UartInterface(Reloadable):
...
+    if platform.system() == 'Linux':
+        DEFAULT_UART_DEV="/dev/ttyACM0"
```
marcan, line 137: This changes the existing behavior. The idea is that on Linux you install the udev rules in `udev/`, which gives you the `/dev/m1n1` symlink. Maybe we can just leave Linux at the default?

marcan (changes requested):

proxyclient/m1n1/proxy.py
aderuelle: Yeah sure

proxyclient/m1n1/proxy.py
aderuelle: Yes fine with me to keep Linux as default

proxyclient/tools/picocom-sec.sh
aderuelle: Nice, I should have figured this out … I will remove this and add some comments about the udev stuff

aderuelle:
> can confirm this works for me out of the box on both a macos and linux host thank you @aderuelle, I carry this on my local envs

thx @martell I removed Linux raw device names as follow up to review comment. 

On Linux this is handled by the[ provided udev rule](https://github.com/AsahiLinux/m1n1/blob/main/udev/80-m1n1.rules) whereas on macOS this is currently hardcoded (there's no udev-like framework AFAIK).

However, we could poll the ioreg database, I've write a [small Python script to do just that.](https://gist.github.com/aderuelle/36454ed691dfceba575411721d8d18b8) I'll have a look into plugging it in someday.

PR #268, opened 2022-12-04
Support concatenated initramfs images

jannau:
To support modules during development with `run_guest_kernel.sh` a cpio archive with matching modules can be created with `make dir-pkg` and `usr/gen_initramfs.sh` in the kernel dir. A specially prepared initramfs can use the modules and make them available to the OS in a similar way as the vendor firmware.


jannau:
@marcan ping. I confuse myself every time I forget to pull this in new branches. Fails here to boot due to missing modules from the initrd.

marcan:
I wanted to try another way but let's just merge this, no point in blocking things further.

PR #274, opened 2023-01-03
firmware: update version arrays to show 12.6.1 and 12.6.2

martell:
dipping my toes in m1n1

I'm curious if there was specific reason to debug on ventura beta 4 ?


marcan:
Did you actually validate that the GPU firmware ABI is identical between 12.6.1 and 13.0beta4? If not, you can't make this change.

I believe @asahilina picked 13.0b4 arbitrarily as whatever was the latest version to demonstrate cross-version compatibility and version conditionals. It is not a supported version, just a test. Nonetheless, please do not remove it from any version arrays unless you can prove it is equivalent to some other version, and definitely don't remove 13.0. These versions having the same iBoot2 version does *not* imply they have the same version of other firmwares, and in that case, that means the version selection logic needs to take into account other factors beyond the iBoot version (which is something I considered but has not been necessary yet).

marcan:
Also the change is outright wrong according to [this](https://www.theiphonewiki.com/wiki/IBoot_(Bootloader)#macOS). 12.6.x uses iBoot-7459.141.1. I don't know where you got that it has the same iBoot version as 13.0. It is highly unlikely for Apple to implement a major iBoot version upgrade between two point releases.

marcan:
Just to clarify what's going here a bit: The primary purpose of the version detection is to forward OS firmware version data to the OS, so it can use the right interfaces. That means we only really care about OS firmware compatibility state, and a about the versions *we* choose to bless as supported. Right now, that is 12.3 for M1 systems and one particular funny version of 12.4 for M2 systems (which was a special launch build and not the regular 12.4).

Other than that, it is useful to populate the system firmware version string in the device tree for informational purposes, but that's not critical since we also have the iBoot1 version (which we can map to a SFR macOS-version later if needed).

So while adding more versions with distinct iBoot version numbers can be handy to provide that info in more cases, we only *have* to add the blessed versions we choose to support, and we should not touch the logic that makes decisions based on those version numbers unless we have verified compatibility across versions.

PR #277, opened 2023-01-21
copy the touchscreen calibration

WhatAmISupposedToPutHere:


src/kboot.c
```diff
@@ -659,6 +659,29 @@ static int dt_set_bluetooth(void)
     return 0;
 }
 
+static int dt_set_multitouch(void)
+{
+    const char *path = fdt_get_alias(dt, "touchscreen0");
```
marcan, line 664: Let's call this `touchbar` (for the same reasons I mentioned when commenting on the firmware).

PR #286, opened 2023-02-14
Prepare SIO firmware data

povik:


src/kboot.c
```diff
@@ -1460,6 +1461,106 @@ static int dt_vram_reserved_region(const char *dcp_alias, const char *disp_alias
...
+    if (ret != 0)
+        bail("DT: could not append phandle '%s.iommu-addresses' property: %d\n", name, ret);
+
+    ret = fdt_appendprop_u64(dt, mem_node, "iommu-addresses", iova);
+    if (ret != 0)
+        bail("DT: could not append iova to '%s.iommu-addresses' property: %d\n", name, ret);
+
+    ret = fdt_appendprop_u64(dt, mem_node, "iommu-addresses", size);
+    if (ret != 0)
+        bail("DT: could not append size to '%s.iommu-addresses' property: %d\n", name, ret);
```
jannau, line 1499: misses to set a `compatible` property for the reserved mem node

src/kboot.c
```diff
@@ -1460,6 +1461,106 @@ static int dt_vram_reserved_region(const char *dcp_alias, const char *disp_alias
...
+    unsigned int num_maps = segments_len / 32;
+
+    for (unsigned i = 0; i < num_maps; i++) {
+        u64 paddr = segments[0];
+        u64 iova = segments[2];
+        u32 size = segments[3];
+        segments += 4;
+
+        char node_name[64];
+        snprintf(node_name, sizeof(node_name), "asc-firmware@%lx", paddr);
```
jannau, line 1553: any reason  not to use `segment-names`?

src/sio.c
```diff
@@ -0,0 +1,259 @@
...
+        goto done;
+    }
+#endif
+
+    if (!sio_num_fwdata++)
+        mapping->iova = *iova = 0x30000;
+    else
+        mapping->iova = *iova = ALIGN_UP((mapping - 1)->iova + (mapping - 1)->size, SZ_16K);
+    mapping->size = ALIGN_UP(size, SZ_4K);
+    mapping->phys = (u64)memalign(SZ_16K, ALIGN_UP(size, SZ_16K));
```
jannau, line 45: Is there maybe already a carveout for sio used by macos? If not allocating this from the top of the usable memory would be a better strategy, see the framebuffer reallocation in `display.c`

src/sio.c
```diff
@@ -0,0 +1,259 @@
...
+#include "types.h"
+#include "utils.h"
+
+// Reuse pages for different data sections, if space allows it
+#define MERGE_SIO_FWDATA
+
+#define SIO_KEY(s)  _SIO_KEY(#s)
+#define _SIO_KEY(s) (((s)[0] << 24) | ((s)[1] << 16) | ((s)[2] << 8) | (s)[3])
+
+#define MAX_FWDATA   8
```
jannau, line 16: this is in conflict with the maximum of 8 `memory-region` in the proposed dt-binding since 2 `memory-region`s are used for firmware text and data

src/kboot.c
```diff
@@ -1460,6 +1461,106 @@ static int dt_vram_reserved_region(const char *dcp_alias, const char *disp_alias
                                    disp_reserved_regions_vram, &region, 1);
 }
 
+static int dt_reserve_device_mem(const char *path, u32 phandle, const char *name, u32 mem_phandle,
+                                 u64 phys, u64 iova, u64 size)
```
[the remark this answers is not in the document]
povik: I think that function may not have existed when I initially wrote this. I will look into it

src/kboot.c
povik: Not really

src/sio.c
povik: Thanks for the tip, will take a look. There isn't a carveout other than one for the main executable sections

src/sio.c
povik: Right

src/sio.c
```diff
@@ -0,0 +1,259 @@
...
+    if (sio_num_fwdata >= MAX_FWDATA)
+        return NULL;
+
+    struct sio_mapping *mapping = &sio_fwdata[sio_num_fwdata];
+
+#ifdef MERGE_SIO_FWDATA
+    if (sio_num_fwdata && ALIGN_UP((mapping - 1)->size, SZ_16K) >= (mapping - 1)->size + size) {
+        mapping--;
+        *iova = mapping->iova + mapping->size;
+        mapping->size = ALIGN_UP(mapping->size + size, SZ_4K);
```
jannau, line 35: not sure if strictly necessary but in the merge case the data is not cleared. easiest workaround would be to always clear the whole 16k page:
`memset64((void *)mapping->phys, 0, ALIGN_UP(mapping->size, SZ_16K));`

povik: Good catch!

povik:
Pushed untested commits to address review (untested). Unaddressed: segment-names and allocating from top of memory

povik:
Changed this from being a draft as it may be ready for inclusion now. FWIW the binding got a r-b from Rob.

src/kboot.c
povik: This I haven't changed

src/kboot.c
```diff
@@ -1596,6 +1661,72 @@ static int dt_set_display(void)
     return dt_vram_reserved_region("dcp", "disp0");
 }
 
+static int dt_set_sio_fwdata(void)
+{
+    const char *path = "/soc/sio";
```
marcan, line 1666: Shouldn't this be an alias? The paths aren't stable (e.g. t600x have the dies)

marcan (commented):
Just a minor comment + whatever wasn't addressed yet.

src/kboot.c
povik: You mean shouldn't we be looking at some aliased path other than `/soc/sio`? AFAIK this will work if `/soc/sio` itself is an alias.

marcan: Aliases are bare labels (like just "sio"), not paths. They point at paths :)

I think just `fdt_path_offset(dt, "sio");` should work, as long as you then add a corresponding `sio` alias (look at `dcp` and `gpu` for examples)

povik:
Rebased (tweaked one commit title; resolved conflict with 91ff36b27)

povik:
This should be OK to merge now. The last blocker was looking up SIO by alias, which has been added now.

PR #302, opened 2023-04-09
Universialize the script

Umeaboy:


Umeaboy:
This commit would make it possible to run it on different shells instead of just using bash.

version.sh
```diff
@@ -14,4 +14,16 @@ else
     version="unknown"
 fi
 
-echo "#define BUILD_TAG \"$version\""
```
marcan, line 17: Why would this not work in every shell?

version.sh
```diff
@@ -14,4 +14,16 @@ else
...
+# Check which shell is running
+if [ -n "${BASH_VERSION:-}" ]; then
+    echo "#define BUILD_TAG \"$version\""
+elif [ -n "${ZSH_VERSION:-}" ]; then
+    printf '#define BUILD_TAG "%s"\n' "$version"
+elif [ -n "${KSH_VERSION:-}" ]; then
+    echo "#define BUILD_TAG=\"$version\""
```
marcan, line 23: This is wrong, that's not valid C.

marcan:
I don't understand why you're trying to detect the shell. If there are any bashisms in use we should replace them with something that works on everything compatible with POSIX sh, there should be no need for shell-specific codepaths.

PR #307, opened 2023-04-22
Support display init with DCP 13.3 firmware

jannau:
minor changes for the 13.3 DCP firmware fixes display init on mac mini (2020) and mac studio (2022)


marcan:
Thanks! :)

PR #314, opened 2023-05-26
clk: Set up LEAP/PDM clocks

povik:


povik:
Not sure how to make clang format do something sane with the clk_data structs, so this is not fully clang-formatted.

marcan:
Try putting a comma after the last element of the arrays. If that doesn't do something sane, just use `// clang-format off` and `// clang-format on` around them :)

marcan:
I think this just needs the clang-format fix to go in.

povik:
Ah, thought I pushed the format fix already. Indeed inserting a comma did the trick.

PR #326, opened 2023-08-29
fix ESR_EC macro

Pricx:
According to the manual, I fix the FEAT_PAuth ESR_EC macro

The manual says:

```
EC == 0b001000
When AArch32 is supported:
Trapped VMRS access, from ID group trap, that is not reported using EC 0b000111. See ISS encoding for an exception from an MCR or MRC access.
EC == 0b001001
When FEAT_PAuth is implemented:
Trapped use of a Pointer authentication instruction because HCR_EL2.API == 0 || SCR_EL3.API == 0.
```


marcan:
Please add the missing signoff.

PR #329, opened 2023-09-03
M2+ desktop dcp/dptx HDMI output support

jannau:
Initialize the HDMI output of M2+ desktop Macs releases in 2023.

- Mac mini (M2/M2 Pro) and Mac studio (M2 Max) should work (tested on both mini models)
- Mac Pro / studio (M2 Ultra) are different but I would expect it to work as well
- only one HDMI port on the Mac Pro works (the one driven by die 1), no idea which
- high res displays might not work if they require different phy parameters

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ LIBFDT_OBJECTS := $(patsubst %,libfdt/%, \
 	fdt_wip.o fdt.o)
 
 DCP_OBJECTS := $(patsubst %,dcp/%, \
+	dpav_ep.o \
 	dptx_phy.o \
 	dptx_port_ep.o \
 	parser.o)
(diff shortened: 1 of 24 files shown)


sequencer:
I'm curious, is this work also help DP Alt for laptops?

niggowai:
I can confirm it working for my `j473` with a `13.5` IPSW.
Unfortunately, linux turns off the dptx again

jannau:
> I can confirm it working for my `j473` with a `13.5` IPSW. Unfortunately, linux turns off the dptx again

yes, linux need further changes not turn the dptx-phy power domain off. You can add `pd_ignore_unused` to the linux kernel command line to avoid that or add `apple,always-on` to the `dptx_ext_phy` power state.

PR #334, opened 2023-09-23
kboot/gpu: Add "no-map" property for reserved memory nodes

jannau:
Without this property u-boot will add reserved memory regions to the EFI memory map with the wrong type. This results in Linux mapping it and thus breaking the intended use in the asahi driver.


marcan:
I think it's fine to just have this in the DT, but I guess being defensive won't hurt.

PR #336, opened 2023-09-30
Parse nvram size from ADT

jannau:
Fixes an issue in `adt_get_reg()` which affect only the range parsing in `spinor`

Can be merged without effects before the device tree updates in https://github.com/AsahiLinux/linux/pull/179

### DIFF
--- a/src/kboot.c
+++ b/src/kboot.c
@@ -729,6 +729,57 @@ static int dt_set_ipd(void)
     return 0;
 }
 
+#define NVRAM_START    0x700000
+#define NVRAM_MAX_SIZE SZ_1M
+
+static int dt_set_nvram(void)
+{
+    int adt_path[8];
+    u64 start, size;
+
+    int anode = adt_path_offset_trace(adt, "/arm-io/spi1/spinor/anvram", adt_path);
+    if (anode < 0) {
+        printf("ADT: nvram partition not found\n");
+        return 0;
+    }
+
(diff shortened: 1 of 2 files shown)


src/kboot.c
```diff
@@ -729,6 +729,57 @@ static int dt_set_ipd(void)
...
+
+    int node = fdt_path_offset(dt, "nvram");
+    if (node < 0) {
+        printf("DT: nvram alias/partition not found\n");
+        return 0;
+    }
+
+    fdt32_t reg[2];
+    fdt32_st(reg + 0, start);
+    fdt32_st(reg + 1, start);
```
nohajc, line 771: I guess this should set `size`. Otherwise the value is unused.
```suggestion
    fdt32_st(reg + 1, size);
```

jannau: thanks for spotting this, fixed. Strange that my dump from j493 using 13.5 seemingly reported the expected mtd0 size.

nohajc: Hmm... and what does `mtdinfo /dev/mtd0` say on your machine?

nohajc: Btw, I still don't know how the "current bank" is picked. I thought each nvram change would increase the counter but now I tried again and the value is still 3 which would suggest the changes are made in place.

Maybe in your case the current bank is within those first 128 kB, so you could see the changes.

PR #344, opened 2023-10-23
bash is not available everywhere

tobhe:
Use /bin/sh instead which should work on most posix platforms. Fixes OpenBSD.


tobhe:
@rowanG077 works for me. I pushed an update.

PR #352, opened 2023-11-11
Add initial M3 values for CPU_START_OFF

dberlin:
I only can easily find the T6031/T6034/T8122 values.
If there is a T6030, the value is not in the same place I found these.


mariogrip:
Seems like there is an T6030, M1 Pro

Screenshot 2024-06-02 at 22 44 03

PR #357, opened 2023-11-22
More support for M3 chips

dberlin:
Now works enough to get proxy + usb working, along with fb/aic/etc.

Trying to start secondary cpus crashes when it hits cpu 4.

I believe this is because we are assuming the primary CPU is always 0, and it is not here, and we are trying to start the CPU we are on right now.


dberlin:
(Just leaving this as a draft until the MMU is not crashing)

dberlin:
Gonna split this pr into a few pieces

PR #363, opened 2023-12-01
Handle case where mem_size_actual is 0

dberlin:
Handle case of no mem_size_actual

On newer machines, mem_size_actual is 0.
In this case, we need to use mem_size and mask to a page boundary.

This fixes MMU crash/hang on m3

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -401,6 +401,13 @@ static void mmu_add_default_mappings(void)
     uint64_t ram_size = cur_boot_args.mem_size + cur_boot_args.phys_base - ram_base;
     ram_size = ALIGN_DOWN(ram_size, 0x4000);
 
+    /* Mem size actual is 0 on some machines, generate a usable value in this case */
+    if (!cur_boot_args.mem_size_actual) {
+        cur_boot_args.mem_size_actual = ALIGN_UP(ram_size, BIT(32));
+        /* Update for python / subsequent stages */
+        memcpy((void *)boot_args_addr, &cur_boot_args, sizeof(cur_boot_args));
+    }
+
     printf("MMU: RAM base: 0x%lx\n", ram_base);
     printf("MMU: Top of normal RAM: 0x%lx\n", ram_base + ram_size);
 


src/memory.c
```diff
@@ -411,7 +412,14 @@ static void mmu_add_default_mappings(void)
...
      */
-    mmu_add_mapping(ram_base, ram_base, cur_boot_args.mem_size_actual, MAIR_IDX_NORMAL, PERM_RWX);
+
+    /*
+     * Newer machines (m3+) do not always have mem_size_actual filled in.  In this case,
+     * use mem size.  Either way, mask it to a page boundary.
+     */
+    u64 mem_size =
+        cur_boot_args.mem_size_actual ? cur_boot_args.mem_size_actual : cur_boot_args.mem_size;
+    mmu_add_mapping(ram_base, ram_base, mem_size & ~(PAGE_SIZE), MAIR_IDX_NORMAL, PERM_RWX);
```
marcan, line 422: This is not correct. `mem_size` is the usable memory size starting at `phys_base`, not at `ram_base`. `ram_size`, calculated above, is the upper limit of normal RAM starting at RAM base, excluding top carveouts. But even if you used that, you'd still break debugging and scripts which rely on upper RAM being mapped (the reason why we explicitly unmap the TZ carveouts is precisely because mapping *those* is dangerous and causes random faults due to speculation, but mapping everything else in high RAM isn't).

What you want to do is the same thing we do above to cook up `ram_base`: just take the wonky value and round to a multiple of 4GiB. In this case, `mem_size_actual = ALIGN_UP(ram_size, BIT(32))`. This should give real physical RAM size for any reasonable configuration.

I'd go ahead and stick it into `cur_boot_args` too, so we don't have to go around fixing other code that relies on it. If macOS doesn't like that field, I'd rather deal with it in the hypervisor boot sequence for macOS. Just do that at the top of the function if `mem_size_actual` is 0 and leave this line alone :)

dberlin:
This should now be correct

jannau:
on j504 (M3 Macbook Pro 14-inch) I see `mem_size_actual` at 0x478 instead of 0x2d8. All data zero so no idea what the additional 416 bytes are used for.

marcan:
Looks like this is an issue with bootargs version/size then?

jannau:
> Looks like this is an issue with bootargs version/size then?

yes, it was. Complicated by the fact that early iboot versions on M3 devices did use revision 2 but used the longer command line length. Missing from #403 is updating `BootArgs` in proxyclient/m1n1/tgtypes.py

mildsunrise:
also missing from #403 is replicating the correction of mem_size_actual (which I understand is for these old iBoot versions) in python side, since we're no longer patching the original bootargs and python sees 0

PR #364, opened 2023-12-01
Support AIC version 3

dberlin:
Support new version of AIC used by M3

### DIFF
--- a/src/aic.c
+++ b/src/aic.c
@@ -31,11 +31,20 @@ static struct aic aic2 = {
         {
             .config = AIC2_IRQ_CFG,
         },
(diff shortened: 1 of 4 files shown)


marcan:
Is the only relevant difference the MAX_DIE field width? Can't we just reuse all the rest of the code and make that conditional / part of the regs struct?

marcan:
SGTM, let's go with that. I'm indifferent about the max dies thing, it's academic until a monster Mac Pro shows up (if ever), and that's guaranteed to need piles of code changes anyway.

dberlin:
This still works :)

PR #365, opened 2023-12-01
Workaround new USB using SPMI instead of I2C

dberlin:
This patch is a workaround for the lack of SPMI support in usb.c

Newer machines use SPMI and a different type C power controller.

This patch still lets us make the USB ports work by skipping power-on and just enabling the phys

### DIFF
--- a/src/usb.c
+++ b/src/usb.c
@@ -7,6 +7,7 @@
 #include "iodev.h"
 #include "malloc.h"
 #include "pmgr.h"
+#include "string.h"
 #include "tps6598x.h"
 #include "types.h"
 #include "usb_dwc3.h"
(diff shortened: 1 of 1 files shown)


dberlin:
Stacked on top of #364

marcan:
Ugh. Well that's going to be fun.

dberlin:
It's actually not too bad - turning on HPM debugging in macos shows
the actual commands that get issued are identical (IE the fourcc and
all), it's just the transport that's different.
So we just have to figure out the command execution protocol and the
rest of the code from the existing i2c driver should work.

I've been working it out in a python experiment, but it's not quite there yet.

PR #367, opened 2023-12-01
Support cpufreq for t6031

dberlin:
This seems to work, and matches the info i can find.

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -194,6 +194,12 @@ static const struct cluster_t t6022_clusters[] = {
     {},
 };
 
+static const struct cluster_t t6031_clusters[] = {
+    {"ECPU0", 0x210e00000, false, 1, 5},
+    {"PCPU0", 0x211e00000, true, 1, 6},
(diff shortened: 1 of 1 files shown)


src/cpufreq.c
```diff
@@ -262,6 +280,8 @@ const struct feat_t *cpufreq_get_features(void)
         case T6021:
         case T6022:
             return t6020_features;
+        case T6031:
+            return t6031_features;
```
aruediger, line 284: same as T6021/T6022?

dberlin: For now, yes.  I assume it will end up slightly different in the end, which is why i didn't reuse the struct.

marcan:
I'm not comfortable merging this until we at least confirm via macOS HV trace that it does the same thing. Getting this wrong has bit us *hard* in the past (M2 core lockups on deep sleep).

dberlin:
That's fine.

dberlin:
Note: I traced enough of the kext now to verify that this is at least as correct as the current one :)

PR #368, opened 2023-12-01
Support boot CPU not being 0.

dberlin:
Right now, we assume the boot cpu is cpu0. That is not true on m3 max,
where it is cpu4.

To figure out which CPU is the boot CPU, we check to see which CPU
is running before we start any other CPUs, and record the MPIDR/idx.

Without this patch, four issues happen on m3 max:

1. We try to start the boot CPU again when we enable SMP, crashing it
2. We skip starting the real CPU 0
3. We start m1n1 again on CPU0 when we _do_ boot it
4. We enable interrupts on CPU0 because we think it's the primary CPU.


dberlin:
I have the fixes for hv/* on the python side coming up, since i had to add
proxy support to retrieve the boot cpu info.
It's good enough to at least crash macos on the first msr ;)

PR #369, opened 2023-12-03
Add support for retrieving boot cpu id and mpidr from proxy

dberlin:
Uses it to add basic support for non-0,0,0 boot cpu in python hypervisor code

### DIFF
--- a/proxyclient/m1n1/hv/__init__.py
+++ b/proxyclient/m1n1/hv/__init__.py
@@ -1558,10 +1558,12 @@ class HV(Reloadable):
             chip_id = self.u.adt["/chosen"].chip_id
             if chip_id in (0x8103, 0x6000, 0x6001, 0x6002):
                 cpu_start = 0x54000 + die * 0x20_0000_0000
-            elif chip_id in (0x8112,):
+            elif chip_id in (0x8112, 0x8122):
                 cpu_start = 0x34000 + die * 0x20_0000_0000
             elif chip_id in (0x6020, 0x6021, 0x6022):
                 cpu_start = 0x28000 + die * 0x20_0000_0000
+            elif chip_id in (0x6031,):
+                cpu_start = 0x88000 + die * 0x20_0000_0000
             else:
                 self.log("CPUSTART unknown for this SoC!")
                 break
@@ -1628,8 +1630,9 @@ class HV(Reloadable):
                 pass
 
         if not self.smp:
+            boot_cpu_name = "cpu" + str(self.p.get_boot_cpu_idx())
             for cpu in list(self.adt["cpus"]):
-                if cpu.name != "cpu0":
+                if cpu.name != boot_cpu_name:
                     print(f"Removing ADT node {cpu._path}")
                     try:
                         del self.adt["cpus"][cpu.name]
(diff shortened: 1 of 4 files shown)


src/proxy.c
```diff
@@ -54,6 +54,12 @@ int proxy_process(ProxyRequest *request, ProxyReply *reply)
         case P_GET_BOOTARGS:
             reply->retval = boot_args_addr;
             break;
+        case P_GET_BOOT_CPU_IDX:
+            reply->retval = boot_cpu_idx;
```
jannau, line 58: I don't think it's a good idea to retrieve something which is only valid after `smp_start_secondaries` is called.

src/proxy.c
```diff
@@ -54,6 +54,12 @@ int proxy_process(ProxyRequest *request, ProxyReply *reply)
...
+            break;
+        case P_GET_BOOT_CPU_MPIDR:
+            reply->retval = boot_cpu_mpidr;
```
jannau, line 61: I don't think anyone has plans to use the non boot CPU for proxy commands so this is equivalent to `u.mrs(MPIDR_EL1)`

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1628,8 +1630,9 @@ def setup_adt(self):
                 pass
 
         if not self.smp:
+            boot_cpu_name = "cpu" + str(self.p.get_boot_cpu_idx())
             for cpu in list(self.adt["cpus"]):
-                if cpu.name != "cpu0":
+                if cpu.name != boot_cpu_name:
```
jannau, line 1635: simpler change without additional proxy commands would use the CPU's ADT node "state" property

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1906,7 +1912,10 @@ def start(self):
         # Does not return
 
         self.started = True
-        self.started_cpus[0] = (0, 0, 0)
+        boot_cpu_name = "cpu" + str(self.p.get_boot_cpu_idx())
+        boot_cpu_adt = self.u.adt["/cpus/" + boot_cpu_name]
+        boot_cpu_mpidr = self.p.get_boot_cpu_mpidr()
+        self.started_cpus[0] = (boot_cpu_adt.die_id, boot_cpu_adt.cluster_id, boot_cpu_mpidr & 0xf)
```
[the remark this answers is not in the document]
dberlin: Will do

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1755,7 +1758,10 @@ def load_raw(self, image, entryoffset=0x800, use_xnu_symbols=False, vmin=0):
 
         print("Setting secondary CPU RVBARs...")
         rvbar = self.entry & ~0xfff
-        for cpu in self.adt["cpus"][1:]:
+        boot_cpu_name = "cpu" + str(self.p.get_boot_cpu_idx())
+        for cpu in self.adt["cpus"]:
+            if cpu.name == boot_cpu_name:
```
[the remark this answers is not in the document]
dberlin: yes, fair enough, i'l fix

src/proxy.c
dberlin: Thoughts on how to approach this?
This was all sort of a hack to get it working.
I could make smp_start_secondaries return this value as an out arg or something?
Or some totally different approach?

src/proxy.c
dberlin: Yes, i'l remove

proxyclient/m1n1/hv/__init__.py
dberlin: Yep.

src/proxy.c
jannau: I would use the `state` property from the ADT. In two cases it is iterating over all CPUs and skipping the boot CPU. The remaining use can be replaced with `[cpu for cpu in self.u.adt["cpus"] if cpu.state == "running"]`.

jannau:
I have the alternate implementation in https://github.com/jannau/m1n1/commits/t8122/ and will create a PR tonight

mildsunrise:
this can be closed now, right?

svenpeter42:
Superseded by #417

PR #373, opened 2023-12-12
docker: install rust toolchain for stage1 builds

davide125:
Install the rust toolchain in the Docker container so it can be used when building with `CHAINLOADING=1` too

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,14 @@
 FROM debian:buster-slim
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && apt-get install -y build-essential bash git locales gcc-aarch64-linux-gnu libc6-dev-arm64-cross device-tree-compiler \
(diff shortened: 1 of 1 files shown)


davide125:
Note: this is installing nightly because that's what the docs say to use (https://github.com/AsahiLinux/docs/wiki/m1n1:User-Guide#build-options), but the build seems to work fine using `stable` these days, so maybe we should use that instead.

Conan-Kudo:
We should definitely use stable toolchains instead of nightly ones.

marcan:
FWIW, we rely on two unstable features (with the same cheat as the Linux kernel to work with stable). One of them is actually deprecated and about to be removed, but it should be trivial to move away from it and to the supported solution (though it will be a flag day for us when it happens). The other is one is one of those things that have been around forever and I have no idea why it isn't stable yet, but shrug.

davide125:
Updated to use stable instead of nightly

PR #374, opened 2023-12-13
Name all Apple registers, at least as they exist on Everest

dberlin:
The internet appears to have discovered the names of all the apple
registers in Everest, along with their encodings.

This fills in the ones we had missing - existing register we named
have not been touched, except for one register (AMX context) that was obviously wrong.

I cannot find a formatter that reproduces the original formatting of the json, if we care.


dberlin:
Whoops, one sec, forgot to filter out the arm ones.

PR #378, opened 2024-01-12
kboot: Add SMBIOS information for U-Boot

kettenis:


kettenis:
U-Boot recently fixed the issue with advertising SMBIOS tables at addresses above 4G that forced me to disable SMBIOS support. Now that it is fixed, let's provide some meaningful information to fill the tables. We could add this directly to the device tree but:
1. Doing so would be more work
2. Would introduce DT schema validation errors since the bindings for this are in the Linux tree (yet)

Personally I don't particularly care about SMBIOS support, but there are Linux tools that expect it to be present. IIRC this does include some bug reporting tools used by Fedora.

chadmed:
I'm not sure if this is a particularly good solution to the problem as we're not really doing anything about the root cause. If certain tools do not support retrieving platform information from the DT, then they are broken on a nontrivial number of computers. Those tools should be fixed, and we shouldn't kludge our way around that brokenness.

Furthermore, having info split across the DT and U-Boot's SMBIOS will subtly break _other_ tools which expect to be parsing information from one or the other, even if it fixes ones that are OF-unaware. For example, this would make KDE Info Centre display less information than it does now, as it will not try and parse the DT if it finds SMBIOS tables first. We're just shifting the problem someplace else by doing this.

Conan-Kudo:
On the other hand, is there a reason we can't make the SMBIOS data more complete? Generally speaking, OF data isn't stable in the same way that SMBIOS data is, and way more tools know how to read SMBIOS than OF data.

src/kboot.c
```diff
@@ -349,6 +349,52 @@ static int dt_set_serial_number(void)
...
+    fdt_setprop_string(dt, node, "compatible", "u-boot,sysinfo-smbios");
+
+    smbios_node = fdt_add_subnode(dt, node, "smbios");
+    if (smbios_node < 0)
+        bail("FDT: failed to add node 'smbios' to  '/smbios'\n");
+
+    node = fdt_add_subnode(dt, smbios_node, "system");
+    if (node < 0)
+        bail("FDT: failed to add node 'system' to  '/smbios'\n");
+    fdt_setprop_string(dt, node, "manufacturer", "Apple");
```
jannau, line 372: Please use `/manufacturer` from the ADT. Value is "Apple Inc." which matches `dmidecode` on Intel Macs. same for baseboard and chassis

src/kboot.c
```diff
@@ -349,6 +349,52 @@ static int dt_set_serial_number(void)
...
+    const char *model = fdt_getprop(dt, fdt_root, "model", &len);
```
jannau, line 373: using `/model` from the ADT would match Intel Macs

src/kboot.c
```diff
@@ -349,6 +349,52 @@ static int dt_set_serial_number(void)
...
+    if (!model)
+        bail("FDT: failed to get model property\n");
+    if (len < 6)
+        fdt_setprop_string(dt, node, "product", "Unknown");
+    else
+        fdt_setprop_string(dt, node, "product", model + 6);
+
```
jannau, line 380: please set "serial" as well. We do that already for the device tree

src/kboot.c
```diff
@@ -349,6 +349,52 @@ static int dt_set_serial_number(void)
...
+    fdt_setprop_string(dt, node, "manufacturer", "Apple");
+    const char *target_type = adt_getprop(adt, 0, "target-type", NULL);
+    if (!target_type)
+        bail("ADT: failed to get target-type property\n");
+    fdt_setprop_string(dt, node, "product", target_type);
+
+    node = fdt_add_subnode(dt, smbios_node, "chassis");
+    if (node < 0)
+        bail("FDT: failed to add node 'chassis' to  '/smbios'\n");
+    fdt_setprop_string(dt, node, "manufacturer", "Apple");
```
jannau, line 393: Please set "chassis-type" as well. Best information to derive it from is "/product/product-name". If it starts with "MacBook" use `laptop` (`0x09`), if it starts with "iMac" use `All in One` (`0x0d`) and otherwise `desktop` (`0x03`). Let's pretend the Mac Pro does not exists.

jannau (commented):
I think we should merge this. `dmidecode` just came up again in https://invent.kde.org/plasma/plasma-welcome/-/merge_requests/213

src/kboot.c
jannau: That's something which is hopefully simple to [change](https://github.com/devicetree-org/dt-schema/pull/162)

src/kboot.c
kettenis: Sure

src/kboot.c
kettenis: Done

src/kboot.c
kettenis: Since "all-in-one" was accepted, I've added code to set the smbios enclosure type based on the "chassis-type" property in the device tree. Which means that property needs to be added to our device trees before this will show up as something else than "unknown" in smbios.

Also note that further u-boot changes are necessary to make this show up correctly anyway as the chassis type is currently only set if CONFIG_GENERATE_SMBIOS_TABLE_VERBOSE=y.

kettenis: BTW, I did add code for "server" just in case that's what folks think is appropripriate for the Mac Pro.

jannau (approved):

svenpeter42 (approved):

PR #381, opened 2024-01-15
aop: 13.5 bring up +  mics + als + generalize afk/epic

eiln:
aop_audio: stream hpai device (high power audio input) microphone samples from aop to admac. output seems to be 32-bit LE floats in groups of 3 (3 mic array)
aop_als: monitor/report red, green, blue, clear, lux, gain etc information from ambient light sensor (ALS)

Tested j293ap 13.5

Also includes afk/epic/rbep generalizations which hopefully don't break DCP


eiln:
nvm this broke dcp. fixing it

proxyclient/m1n1/fw/afk/rbep.py
```diff
@@ -50,19 +52,24 @@ class AFKEP_Shutdown(AFKEPMessage):
 class AFKEP_Shutdown_Ack(AFKEPMessage):
     TYPE = 63, 48, Constant(0xc1)
 
-
-class AFKError(Exception):
-    pass
+# TODO no one else but AOP uses AFKRingBuf right?
```
jannau, line 55: DCP uses AFKRingBuf on some endpoints via `EPICEndpoint`. Used in experiments/dcpext_iboot.py

eiln: New commits don't break DCP experiments. DCP/AOP AFKRingbuf are too different to share (e.g. different EPIC header, different ring size, AOP also writes ptrs to DART and etc), everything else is fine though. I think we can move afk to soc/apple/ and just have fn pointers for the rbep r/w calls

eiln:
@jannau  I generalized AFK/EPIC code to both DCP and AOP. I feel like I should send these in a separate series but it obviously depends on the above. I think marcan's on holiday. From my tests (dcp/dcp_iboot; I can't test dptx/dcpext) it doesn't break DCP but I'd like to get your ack

pbasista:
Hi, could someone familiar with Asahi Linux and Apple audio help to move this PR or more generally the microphone support forward?

Issue #393, opened 2024-06-03
m1n1 make: src/utils_asm.S:115:5: error: instruction requires: fp-armv8 


vasvasco:
I have an m3 apple and I am trying to make m1n1 from the GitHub repository https://github.com/AsahiLinux/m1n1.git

I did install the llvm clang and other chaintools from Homebrew but I get the following error:

src/utils_asm.S:115:5: error: instruction requires: fp-armv8

PR #394, opened 2024-06-06
Makefile: Fix build when using clang as AS

ayrtonm:
Unlike gcc, clang enforces -mgeneral-regs-only when used as the assembler so this removes that flag to fix the error in #393. This is only needed for utils_asm.S which has code saving/restoring SIMD state, but this commit applies it to all asm files for simplicity since code for fp regs shouldn't be generated unless they're explicitly used anyway.


PaulCharlton:
@marcan ?

waterdragon78:
I needed this too.

PR #399, opened 2024-08-10
Add ADT and m1n1 stage 2 log as MTD phram reserved memory nodes

jannau:
Requires for easy access `CONFIG_MTD_PHRAM` but I guess can be still accesses through /dev/mem on systems missing this.

The additional IODEV for the log console is a little bit ugly since it can't be added before `IODEV_USB0` to keep backwards compatibility with python m1n1 but it shouldn't be handled like the USB IODEVs.

### DIFF
--- a/src/iodev.c
+++ b/src/iodev.c
@@ -18,17 +18,19 @@
 
 extern struct iodev iodev_uart;
 extern struct iodev iodev_fb;
+extern struct iodev iodev_log;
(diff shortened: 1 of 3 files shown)


jannau:
`hexdump -C /dev/mtd/by-name/adt` and `cat /dev/mtd/by-name/m1n1_stage2.log` produce the expected results

marcan:
Nice!

PR #402, opened 2024-09-09
T6030 (M3 Pro) changes

emxll:
Hey, here are some changes I made in order to get m1n1 to work (better, not perfectly though) on my chip.

A couple of notes:

I don't know how much work has been done on AIC3, but as far as I can tell, the interface is identical to (or a superset of) that of AIC2, with only the initial CFG offset being different. I've written `aic3_init` to be more inline with what the official driver does (mainly pulling everything from the adt (edt?) instead of hardcoding it like m1n1 does) but `aic2_init` is able to initialize everything just fine. 

`CMDLINE_LENGTH` is 1024 (not 608) on my system/version of macOS but I'm not quite sure what the plan is for addressing this issue which is why I'm not submitting a patch. 

The RTBuddy/`__OSLOG` (causing illegal address panics in `phystokv`) fix for booting macOS isn't enough for my kernelcache. I'll investigate this more if I can find the time.

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -19,8 +19,13 @@
 #define CLUSTER_PSTATE_DESIRED2               GENMASK(15, 12)
 #define CLUSTER_PSTATE_APSC_BUSY              BIT(7)
 #define CLUSTER_PSTATE_DESIRED1               GENMASK(4, 0)
+#define CLUSTER_PSTATE_DESIRED1_S5L8960X      GENMASK(24, 22)
 
-#define CLUSTER_SWITCH_TIMEOUT 100
+#define PMGR_VOLTAGE_CTL_OFF_S5L8960X 0x20c00
+#define PMGR_VOLTAGE_CTL_OFF_T7000    0x23000
+#define PMGR_VOLTAGE_CTL_OFF_S8000    0xa0000
+
+#define CLUSTER_SWITCH_TIMEOUT 400
 
 struct cluster_t {
     const char *name;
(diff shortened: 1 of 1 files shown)


marcan:
I think you'll want to look at the PRs by @dberlin (e.g. #364) since he did most of the M3 m1n1 bringup already, it's just not merged yet because we didn't get around to pushing forward with that since priorities shifted elsewhere.

emxll:
Once those changes have been merged i'll rebase and see if I'm still missing anything.

flatz:
@emxll @marcan I was able to fix **__OSLOG** panics by tweaking **segment_names/segment_ranges** cleaning code on Python's side:

adt.py:
```python
def remove_os_log_segments(adt, name):
    key = f"/arm-io/{name}"
    if key in adt:
        entry = adt[key]
        try:
            segment_names = entry.segment_names.split(";")
            idx = segment_names.index("__OS_LOG")
            del segment_names[idx]
            entry.segment_names = ";".join(segment_names)
            entry.segment_ranges = entry.segment_ranges[:idx * 32] + entry.segment_ranges[(idx + 1) * 32:] # struct adt_segment_ranges
            return True
        except:
            pass
    return False

```
Change block within `for name in ("mtp", "aop"):`:
**__init__.py:**
```python
        for name in ("mtp", "aop", "dcp", "dcpext"):
            if name in self.adt["/arm-io"]:
                key = f"/arm-io/{name}"
                remove_os_log_segments(self.adt, name)
                remove_os_log_segments(self.adt, f"{name}/iop-{name}-nub")
```

**chainload.py:**
```python
for name in ("mtp", "aop", "dcp", "dcpext"):
    if name in u.adt["/arm-io"]:
        key = f"/arm-io/{name}"
        remove_os_log_segments(u.adt, name)
        remove_os_log_segments(u.adt, f"{name}/iop-{name}-nub")
```

Also older cleaning code has a bug, **OSLOG** segment could be in the middle and not in the end as it may expect here: https://github.com/AsahiLinux/m1n1/blob/9d445d829170edbff6e703705b0c82210024fa6a/proxyclient/m1n1/hv/__init__.py#L1741

dberlin:
> Once those changes have been merged i'll rebase and see if I'm still missing anything.

FWIW:
I'm around and don't have a lot of time to code (but will in a few months, actually).

At the point i uploaded all my PR's, i could boot and run osx under m1n1.

I still have notes on things if you run into any issues.

mildsunrise:
although most of the patches here are no longer needed, some are still of value.
@emxll could you add the missing sign offs when you have some time?

PR #403, opened 2024-09-23
A7-A11 SoC support

asdfugil:
Add support for A7-A11 SoCs found in various iPhones, iPads, iPod touches and Apple TVs.

### DIFF
--- a/m1n1-raw.ld
+++ b/m1n1-raw.ld
@@ -1,6 +1,7 @@
 ENTRY(_start)
 
 _stack_size = 0x20000;
+_stack_size_el3 = 0x8000;
 
 /* We are actually relocatable */
 . = 0;
@@ -71,6 +72,12 @@ SECTIONS {
         *(COMMON)
         _bss_end = .;
     } : data
+    .stack_el3 : ALIGN(0x4000) {
+        PROVIDE(_stack_top_el3 = .);
+        . += _stack_size_el3 - 8;
+        QUAD(0x5176694b43415453);
+        PROVIDE(_stack_bot_el3 = .);
+    }
     .stack : ALIGN(0x4000) {
         PROVIDE(_stack_top = .);
         . += _stack_size - 8;
--- a/src/chickens.c
+++ b/src/chickens.c
@@ -70,7 +70,7 @@ void init_t6031_sawtooth(void);
 void init_t6031_everest(int rev);
 
 bool cpufeat_actlr_el2, cpufeat_fast_ipi, cpufeat_mmu_sprr;
-bool cpufeat_global_sleep;
+bool cpufeat_global_sleep, cpufeat_workaround_cyclone_cache;
 
 const char *init_cpu(void)
 {
@@ -244,9 +244,11 @@ const char *init_cpu(void)
 
     if (part >= MIDR_PART_T8010_2_HURRICANE)
         cpufeat_global_sleep = true;
-    else
+    else {
         /* Disable deep sleep */
         reg_clr(SYS_IMP_APL_ACC_CFG, ACC_CFG_DEEP_SLEEP);
+        cpufeat_workaround_cyclone_cache = true;
+    }
 
     /* Unmask external IRQs, set WFI mode to up (2) */
     reg_mask(SYS_IMP_APL_CYC_OVRD,
--- a/src/chickens.c
+++ b/src/chickens.c
@@ -69,7 +69,7 @@ void init_t6021_avalanche(int rev);
 void init_t6031_sawtooth(void);
 void init_t6031_everest(int rev);
 
-bool cpufeat_actlr_el2, cpufeat_fast_ipi;
+bool cpufeat_actlr_el2, cpufeat_fast_ipi, cpufeat_mmu_sprr;
 
 const char *init_cpu(void)
 {
@@ -230,10 +230,16 @@ const char *init_cpu(void)
         sysop("isb");
 
         msr(SYS_IMP_APL_AMX_CTX_EL1, core);
+
+        /* T8030 SPRR is different */
+        cpufeat_mmu_sprr = true;
     }
 
     if (part >= MIDR_PART_T8030_LIGHTNING)
         msr(SYS_IMP_APL_AMX_CTL_EL1, 0x100);
+
+    if (part >= MIDR_PART_T8015_MONSOON)
+        cpufeat_fast_ipi = true;
 
     /* Unmask external IRQs, set WFI mode to up (2) */
     reg_mask(SYS_IMP_APL_CYC_OVRD,
--- a/src/chickens.c
+++ b/src/chickens.c
@@ -69,7 +69,7 @@ void init_t6021_avalanche(int rev);
 void init_t6031_sawtooth(void);
 void init_t6031_everest(int rev);
 
-bool cpufeat_actlr_el2;
+bool cpufeat_actlr_el2, cpufeat_fast_ipi;
 
 const char *init_cpu(void)
 {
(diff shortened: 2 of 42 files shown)


src/chickens.c
```diff
@@ -147,24 +221,27 @@ const char *init_cpu(void)
...
-    // Enable IRQs (at least necessary on t600x)
-    // XXX 0 causes pathological behavior in EL1, 2 works.
-    msr(SYS_IMP_APL_SIQ_CFG_EL1, 2);
+        // Unknown, related to SMP?
+        msr(s3_4_c15_c5_0, core);
+    }
 
-    sysop("isb");
+    if (part >= MIDR_PART_T8030_LIGHTNING)
+        msr(SYS_IMP_APL_AMX_CTL_EL1, 0x100);
```
asdfugil, line 238: A13 is checked with project triangulation ppl bypass with a dirty cache line. Not very useful though.

src/cpu_regs.h
```diff
@@ -144,11 +148,13 @@
 #define EHID1_EN_LFSR_STALL_RS6                  BIT(62)
 #define EHID1_EN_LFSR                            BIT(63)
 
-#define SYS_IMP_APL_HID3                  sys_reg(3, 0, 15, 3, 0)
-#define HID3_DISABLE_ARBITER_FIX_BIF_CRD  BIT(44)
-#define HID3_DEV_PCIE_THROTTLE_LIMIT_MASK GENMASK(62, 57)
-#define HID3_DEV_PCIE_THROTTLE_LIMIT(x)   ((ULONG(x)) << 57)
-#define HID3_DEV_PCIE_THROTTLE_ENABLE     BIT(63)
+#define SYS_IMP_APL_HID3                                   sys_reg(3, 0, 15, 3, 0)
+#define HID3_DisDcZvaCmdOnly                               BIT(25)
```
marcan, line 152: Please reformat this to uppercase with underscores like the other flags (same with all the others).

m1n1-raw.ld
```diff
@@ -77,6 +78,12 @@ SECTIONS {
         QUAD(0x544f424b43415453);
         PROVIDE(_stack_bot = .);
     } :data
+    .stack_el3 : ALIGN(0x4000) {
+        PROVIDE(_stack_top_el3 = .);
+        . += _stack_size_el3 - 8;
+        QUAD(0x544f424b43415453);
```
marcan, line 84: This is a magic number to find the end of the binary. Please move this new stack before the regular stack and remove the magic.

src/exception_asm.S
```diff
@@ -46,7 +43,6 @@ _v_sp0_fiq:
 .globl _v_sp0_serr
 .type _v_sp0_serr, @function
 _v_sp0_serr:
-    msr pan, #0
```
marcan, line 49: This has to be the first thing before any stack/mem accesses, it's here and duplicated for a reason. It won't work if you move it like this.

src/utils.c
```diff
@@ -181,6 +185,33 @@ bool is_heap(void *addr)
     return p > top_of_kernel_data && p < top_of_ram;
 }
 
+bool is_idevice(void)
+{
+    return chip_id == 0x8960 || chip_id == 0x7000 || chip_id == 0x7001 || chip_id == 0x8000 ||
```
marcan, line 190: This seems like bad naming. There are iDevices with the same chips as Macs. If this is about "old" devices / before a specific cutoff, we should use another function name. If it's about iDevices (e.g. it would hypothetically apply to M1 iDevices) then the test needs to be something else than chip IDs.

src/utils.c
```diff
@@ -181,6 +185,33 @@ bool is_heap(void *addr)
...
+}
+
+bool supports_arch_retention(void)
+{
+    return mrs(AIDR_EL1) & AIDR_EL1_ARCH_RETENTION;
+}
+
+bool supports_fipi(void)
+{
+    return (!is_idevice() || chip_id == 0x8015);
```
marcan, line 202: Probably just test the boot CPU part number for this? See the test for architectural ACTLR_EL12 that got added recently, we can have more flags like that.

src/startup.c
```diff
@@ -63,11 +69,53 @@ void dump_boot_args(struct boot_args *ba)
...
+
+    if (node < 0) {
+        printf("ADT: no /chosen found\n");
+        return;
+    }
+
+    u32 len;
+    const char *p = adt_getprop(adt, node, "firmware-version", &len);
+    u16 version = ba->revision;
+    if (strncmp(p, "iBoot-", 6) == 0) {
```
marcan, line 82: We just added common code for this kind of test (for the Sequoia workaround), please use that. See firmware.c.

src/exception.c
```diff
@@ -24,13 +24,13 @@ volatile int exc_count = 0;
...
-    [0x08] = "EL2t", //
-    [0x09] = "EL2h", //
-};
+static char *m_table[0x10] = {[0x00] = "EL0t", //
+                              [0x04] = "EL1t", //
+                              [0x05] = "EL1h", //
+                              [0x08] = "EL2t", //
+                              [0x09] = "EL2h", //
+                              [0x0c] = "EL3t", //
+                              [0x0d] = "EL3h"};
```
marcan, line 33: Comma at the end here will keep the format sane.

marcan (changes requested):

src/exception_asm.S
asdfugil: Well then this tricky, guess I will need to do runtime code patching during init.

src/utils.c
asdfugil: This is used for both purposes so this function needs to be split.

src/startup.c
asdfugil: I will see but I think I will need a separate fw_version_info because this is not macOS.

asdfugil: Also in my case there's no guarantee of what versions the device may be on so I may have to add absolutely every release and beta...

asdfugil: I think I will add a new function to the common code to directly compare iboot version.

asdfugil:
Changes:
- Register names are now upper cased
- EL3 stack is now before main stack, and does not have `STACKBOT` magic anymore.
- `is_idevice()` is replaced by `is_mac`, `cpufeat_fast_ipi` and `cpufeat_mmu_sprr`, or removed entirely when not needed.
  - `supports_fipi()` removed as well
- Use macros like `S5L8960X` for all `chip_id` comparisons 
- PAN in exception handler now remains unchanged in code but get patched to NOP during init on non-PAN devices.
- Use common code to parse iBoot versions
- Add comma to EL list to keep format sane

src/startup.c
marcan: The iBoot version compare code does not rely on the firmware version table being exhaustive, it directly compares the iBoot version string. You only have to add the specific *threshold* versions to the table you want to test for, then anything in the range you specify will be considered as testing positive. I added it specifically because we need to test the iBoot1 version range (which we do not control) while we do control the subset of iBoot2 verisons we use.

asdfugil: This particular piece of code is ran early, before `firmware_init()`, so this code still needs to parse iBoot version itself.

asdfugil:
Changes v3:
- Fixed PMGR on A7-A11 SoCs. These SoC uses a struct layout that has 8-bit device IDs. Seems like at some point between A11 and M1 Apple needs more than 255 devices and switched to 16-bit device IDs.
- Apple TV HD debug uart is uart6, so account for that
- Add A7-A9 cache errata workaround
- Introduce `el3_call()`
- Fixed `el0_call()` and `el1_call()` when running in EL1

asdfugil:
(... for the proxyclient I don't have anything that looks remotely okay it's all ugly hacks)

marcan (approved):

marcan:
Let's merge, if there are any issues we'll fix as we go along. Thanks!

PR #404, opened 2024-09-23
Fix fallout/leftovers from architectural ACTLR_EL12 support on M2+

jannau:


jannau:
updated to include #405

PR #408, opened 2024-09-26
m1n1: Fix encoding for several ARM special registers

flatz:
Fix register encoding numbers for those of them that have additional index parameter (e.g. DBGWCR0_EL1, DBGWCR1_EL1)


marcan:
This file is generated from ARM register specifications using `reg2json.py`. If you want to make changes you need to change the script, the JSON should never be edited manually.

PR #413, opened 2024-10-27
improvements to ADT dumper

mildsunrise:
improves the output from adt.py in several ways.
some are needed to make diffs possible (most importantly, phandles are formatted as the path to the target node) and some are just for better legibility. see the individual commit messages for details.

(draft since the last commit is dirty code...)


mildsunrise:
I also put this in my `.git/config`:

```
[diff "adt"]
    textconv = "python -m proxyclient.m1n1.adt --sort-keys --sort-nodes"
    xfuncname = "^ *[a-zA-Z0-9_-]+ \\{$"
```

and this in `.gitattributes`:

```
*.adt   diff=adt
```

so that I am able to diff blobs directly, e.g. `git diff --no-index DeviceTree.j51{6,4}map.adt`

PR #417, opened 2024-10-30
Support arbitrary boot CPU core

jannau:
M3 devices use the first CPU core in the first performance core cluster for booting while the efficiency core cluster remains the first cluster. Use the CPU core's state property to determine the boot CPU.

Supersedes #369

Tested on M3 MacBook Pro (2023, 14-inch).

### DIFF
--- a/proxyclient/m1n1/hv/__init__.py
+++ b/proxyclient/m1n1/hv/__init__.py
@@ -1562,10 +1562,12 @@ class HV(Reloadable):
             chip_id = self.u.adt["/chosen"].chip_id
             if chip_id in (0x8103, 0x6000, 0x6001, 0x6002):
                 cpu_start = 0x54000 + die * 0x20_0000_0000
-            elif chip_id in (0x8112,):
+            elif chip_id in (0x8112, 0x8122):
                 cpu_start = 0x34000 + die * 0x20_0000_0000
             elif chip_id in (0x6020, 0x6021, 0x6022):
                 cpu_start = 0x28000 + die * 0x20_0000_0000
+            elif chip_id in (0x6031,):
+                cpu_start = 0x88000 + die * 0x20_0000_0000
             else:
                 self.log("CPUSTART unknown for this SoC!")
                 break
--- a/proxyclient/m1n1/hv/__init__.py
+++ b/proxyclient/m1n1/hv/__init__.py
@@ -90,7 +90,7 @@ class HV(Reloadable):
         self.sym_offset = 0
         self.symbols = []
         self.symbol_dict = {}
-        self.sysreg = {0: {}}
+        self.sysreg = {}
         self.novm = False
         self._in_handler = False
         self._sigint_pending = False
@@ -1633,7 +1633,7 @@ class HV(Reloadable):
 
         if not self.smp:
             for cpu in list(self.adt["cpus"]):
-                if cpu.name != "cpu0":
+                if cpu.state != "running":
                     print(f"Removing ADT node {cpu._path}")
                     try:
                         del self.adt["cpus"][cpu.name]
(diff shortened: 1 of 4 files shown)


src/startup.c
```diff
@@ -156,6 +152,26 @@ void _start_c(void *boot_args, void *base)
...
+#ifndef BRINGUP
+    int node = adt_path_offset(adt, "/cpus");
+    if (node >= 0) {
+        ADT_FOREACH_CHILD(adt, node)
+        {
+            const char *state = adt_getprop(adt, node, "state", NULL);
+            if (!state)
+                continue;
+            if (strcmp(state, "running") == 0)
+                if (ADT_GETPROP(adt, node, "cpu-id", &cpu_id) == sizeof(cpu_id))
```
asdfugil, line 164: Please add a fallback to `reg` here, iPhone 5s iOS 10.3.3 ADT does not have `cpu-id`.

asdfugil (changes requested):

src/startup.c
jannau: `cpu_id` is initialized to 0 and it remains untouched if the the CPU nodes have no `cpu-id` property.

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1562,10 +1562,12 @@ def cpustart_wh(base, off, data, width):
             chip_id = self.u.adt["/chosen"].chip_id
             if chip_id in (0x8103, 0x6000, 0x6001, 0x6002):
                 cpu_start = 0x54000 + die * 0x20_0000_0000
-            elif chip_id in (0x8112,):
+            elif chip_id in (0x8112, 0x8122):
                 cpu_start = 0x34000 + die * 0x20_0000_0000
             elif chip_id in (0x6020, 0x6021, 0x6022):
                 cpu_start = 0x28000 + die * 0x20_0000_0000
+            elif chip_id in (0x6030, 0x6031, 0x6034,):
+                cpu_start = 0x88000 + die * 0x20_0000_0000
```
mildsunrise, line 1570: this is incorrect, Pro has a different CPU start offset than Max. I'd suggest leaving T6030 out of this PR

mildsunrise:
thanks for putting up a PR for this! your patches match with mine except for two extra things, they're mostly nits:

``` diff
diff --git a/proxyclient/m1n1/hv/__init__.py b/proxyclient/m1n1/hv/__init__.py
index a96c793..4458e14 100644
--- a/proxyclient/m1n1/hv/__init__.py
+++ b/proxyclient/m1n1/hv/__init__.py
@@ -90,7 +90,7 @@ class HV(Reloadable):
         self.sym_offset = 0
         self.symbols = []
         self.symbol_dict = {}
-        self.sysreg = {0: {}}
+        self.sysreg = {}
         self.novm = False
         self._in_handler = False
         self._sigint_pending = False
diff --git a/src/hv_exc.c b/src/hv_exc.c
index 3533802..803f14f 100644
--- a/src/hv_exc.c
+++ b/src/hv_exc.c
@@ -522,7 +522,7 @@ void hv_exc_fiq(struct exc_info *ctx)
 
     int interruptible_cpu = hv_pinned_cpu;
     if (interruptible_cpu == -1)
-        interruptible_cpu = 0;
+        interruptible_cpu = boot_cpu_idx;
 
     if (smp_id() != interruptible_cpu && !(mrs(ISR_EL1) & 0x40) && hv_want_cpu == -1) {
         // Non-interruptible CPU and it was just a timer tick (or spurious), so just update FIQs
```

jannau:
Thanks for the review. I've removed the chip IDs not tested by Daniel or me and pulled the sysreg change in `proxyclient/m1n1/hv/__init__.py` into the existing commit.
Please add the `interruptible_cpu` fix as your own commit to #418

marcan:
I squashed in the interruptible_cpu change since it's trivial.

PR #419, opened 2024-11-03
pmgr: Fix false positive u8 pmgr ID detection on modern SoCs

jannau:
The id1 field used on oder SoCs is not constant 0 for all devices on t8122. Instead just compare the id1 fields of the two first devices. Fixes "pmgr: Failed to find parent #36 for IOA0\n" errors in pmgr_init().

Fixes: fb59dc03ca9a ("pmgr: Fix struct for older SoCs")

### DIFF
--- a/src/pmgr.c
+++ b/src/pmgr.c
@@ -374,12 +374,14 @@ int pmgr_init(void)
 
     printf("pmgr: Cleaning up device states...\n");
 
+    // detect whether u8 or u16 PMGR IDs are used by comparing the IDs of the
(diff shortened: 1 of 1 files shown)


jannau:
Cc: @asdfugil does this still work on A7 - A11?

asdfugil:
Works on A7-A11, T2, so:

Reviewed-by: Nick Chan <towinchenmi@gmail.com>

asdfugil (approved):

PR #420, opened 2024-11-10
kboot: Pass through SEPFW and boot object manifests

WhatAmISupposedToPutHere:


src/kboot.c
```diff
@@ -1882,6 +1882,53 @@ static int dt_set_display(void)
...
+        bail("ADT: /chosen/memory-map not found \n");
+
+    u64 phys_map[2];
+    size_t ret = ADT_GETPROP_ARRAY(adt, anode_mmap, "SEPFW", phys_map);
+    if (ret != sizeof(phys_map))
+        bail("ADT: could not get sepfw memory\n");
+
+    const char *node_name = "sep-firmware";
+    int mem_node =
+        dt_get_or_add_reserved_mem(node_name, "apple,asc-mem", false, phys_map[0], phys_map[1]);
```
marcan, line 1902: I think we use asc-mem for memory mapped directly to an ASC (like actual firmware that is pre-loaded). Is SEPFW like that (does it get mapped directly to SEP?) or is it just a blob that the driver does something with? If the latter, it probably should have some other compatible.

src/kboot.c
```diff
@@ -2436,6 +2483,8 @@ int kboot_prepare_dt(void *fdt)
         return -1;
     if (dt_set_multitouch())
         return -1;
+    if (dt_set_sep())
```
marcan, line 2486: This is going to blow up on DTs without the SEP alias. That needs to be handled gracefully for backwards compat.

src/kboot.c
```diff
@@ -1882,6 +1882,53 @@ static int dt_set_display(void)
     return dt_vram_reserved_region(disp_cfg->dcp_alias, "disp0");
 }
 
+static int dt_set_sep(void)
+{
+    const char *path = fdt_get_alias(dt, "sep");
+    if (path == NULL)
+        bail("FDT: sep alias not found in devtree\n");
```
marcan, line 1889: (re last comment) probably do an info print and just return 0 here.

src/kboot.c
WhatAmISupposedToPutHere: Mapped directly to sep, and then copied by sep bootrom into protected memory.

marcan: Fair enough then.

PR #425, opened 2024-12-01
Add support for booting on T8012 (T2) SoC

asdfugil:
This PR contains the changes required to boot the kernel on T2 per this LKML submission:

https://lore.kernel.org/asahi/20241201161942.36027-1-towinchenmi@gmail.com/T/#t

Overall T2 is similar to A10, and this takes care of the remaining:

- The ultra-narrow touch bar screen, or no screen at all
- Some memory channels being unused on some devices
- Some ADT oddities
- Takes the memory intended for SSD cache and allow the kernel to use it (If a driver for booting x86 ever end up getting made that memory has to be described anyways)

----
P.S. T2 Boots on the P-core in Apple Fusion Architecture


src/kboot.c
```diff
@@ -396,6 +397,20 @@ static int dt_set_memory(void)
         }
     }
 
+    if (chip_id == T8012) {
+        u64 cache_min = ALIGN_UP(dram_max, BIT(29));
+        u64 cache_max = dram_base + dram_size;
+
+        /* Set x86 SSD Cache as OS memory, any potential driver would need the memory described
```
marcan, line 404: This won't work with an actual driver, it just abuses this region as main memory. If a driver were to use this, it would have to be described as reserved-memory (which automatically adds it to the list if mapped).

If you intend to ever actually support ANS here, then this probably needs to be kept reserved.

src/kboot.c
```diff
@@ -396,6 +397,20 @@ static int dt_set_memory(void)
...
+         * anyways */
+
+        memreg[num_regions].start = cpu_to_fdt64(cache_min);
```
marcan, line 407: Missing bounds check.

marcan (changes requested):

src/kboot.c
asdfugil: I think I will not describe this memory region for now, given that it might need new compatibles and stuff and that that part of the hardware is not well understood. In this case, this memory region only needs to be concerned while correcting `mem_size_actual`.

asdfugil:
Changes:

- Drop description of the SSD cache memory region in FDT setup.

src/kboot.c
```diff
@@ -2383,6 +2384,37 @@ static int dt_transfer_virtios(void)
...
+        printf("FDT: Failed to find pmgr node\n");
+        return 0;
+    }
+
+    int dcs2_node = fdt_path_offset(dt, "/soc/power-management@20e000000/power-controller@80258");
+    if (dcs2_node < 0)
+        bail("FDT: failed to find ps_dcs2 node\n");
+
+    /* Allow failure */
+    fdt_delprop(dt, dcs2_node, "apple,always-on");
```
svenpeter42, line 2406: Any reason this isn't just done in the devicetree?

src/startup.c
```diff
@@ -134,7 +134,23 @@ void dump_boot_args(struct boot_args *ba)
             break;
     }
     if (!mem_size_actual) {
-        mem_size_actual = ALIGN_UP(ba->phys_base + ba->mem_size - 0x800000000, BIT(30));
+        if (chip_id == T8012) {
+            int anode = adt_path_offset(adt, "/arm-io/mcc");
+
+            /*
+             * Lower 512 MB intented for OS use, upper 512 or 1536 MB is some sort
```
svenpeter42, line 141: Aren't we the OS? Or is there something else running there? And I'm not sure how this comments related to the code below which just computes the memory size.

The SSD cache is also present on >M1 fwiw, it's transparently handled by ANS firmware there and we can't even access it from the AP.

svenpeter42 (changes requested):

src/kboot.c
```diff
@@ -207,7 +207,8 @@ static int dt_set_chosen(void)
         printf("FDT: initrd at %p size 0x%lx\n", initrd_start, initrd_size);
     }
 
-    if (cur_boot_args.video.base) {
+    if (cur_boot_args.video.base &&
+        (adt_path_offset(adt, "/arm-io/disp0") > 0 || (chip_id != T8012 && chip_id != T7000))) {
```
svenpeter42, line 211: Could you also add a comment about this condition? Looks somewhat confusing to me

src/kboot.c
asdfugil: It can't be done in the device tree because the amount of memory and the amount of used memory channels depends on the storage configuration of the host Mac.

src/startup.c
asdfugil: The comment help explain why the amount of reserved memory is so big. The upper memory region has something to do with exporting the ANS2 storage to the x86 CPU (which, on a T2 machine, is more like a big fat coprocessor of the T2).

src/kboot.c
svenpeter42: Oh, I totally missed the mem_size_actual check. I'd add a comment to this function then to explain what's going on here.

src/startup.c
svenpeter42: I think that upper memory is probably just Apple's variant of host memory buffers, i.e. using the main RAM as cache for the nvme drive. I don't even think that will need any driver support as long as you make sure not to clobber that memory region. When I briefly looked at how NVMe worked on T2 macs it seemed like it's just exporting the same MMIO interface we directly use on >M1 to the x86 CPU.

What I'm confused about here is that this comment talks about two reserved memory regions at different locations (lower and upper) but then the code just computes mem_size_actual.

src/kboot.c
asdfugil: Sure, this condition is that there exist T8012 and T7000 devices with no display at all.

svenpeter42: Okay, but do they still have a /arm-io/disp0 node then?

asdfugil: That node doesn't exist in their ADT.

svenpeter42: okay, so please take a step back and explain what you are trying to do here because this condition doesn't make much sense to me.

Right now we will init the framebuffer if
cur_boot_args.video.base is not null and either
    /arm-io/disp0 exists
        OR
    the chip is not T8012 and also not T7000

asdfugil: This condition is to ensure on chips that definitely have a display attached, the framebuffer will be initialized if passed by iBoot, but now I am aware that some devices uses `/arm-io/dispext4` and this would actually break that, so how about:

Init framebuffer if base exists and (either chip is S5L8960X or the dimensions are not 640x1136)
(Would miss 640x1136 external display (?) does those exist (?))

svenpeter42: Does it even hurt to initialize the framebuffer when no display is connected?

svenpeter42: So why not just check if cur_boot_args.video.base is set and /chosen/framebuffer exists then?

asdfugil:
Changes: 
- Do not require `/chosen/framebuffer` to exist when booting Linux. Instead, use its existence to determine whether the device have a display or is capable to have a display connected. Skip framebuffer initialization if it does not exist
- Add comment about computing mem_size_actual for T2
- Add comment about disabling apple,always-on for unused memory channels in T2 pmgr.

PR #428, opened 2024-12-18
Fix up desktop display init regression

chadmed:
Commit 869d2ae ~inadvertently dropped the call to `get_device_info()` and also~ prevented DCP from initialising on Mac desktops. This series fixes ~both of those things~.

I'm not entirely happy with the way we're checking for a DCP, it's a bit ugly.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -51,7 +51,7 @@ void get_device_info(void)
         printf("  Target: %s\n", target);
 
     is_mac = !!strstr(model, "Mac");
-    has_dcp = adt_path_offset(adt, "/arm-io/dcp") > 0;
+    has_dcp = !!adt_path_offset(adt, "/arm-io/dcp") || !!adt_path_offset(adt, "/arm-io/dcpext0");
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -51,7 +51,7 @@ void get_device_info(void)
         printf("  Target: %s\n", target);
 
     is_mac = !!strstr(model, "Mac");
-    has_dcp = adt_path_offset(adt, "/arm-io/dcp") > 0;
+    has_dcp = !!adt_path_offset(adt, "/arm-io/dcp") || !!adt_path_offset(adt, "/arm-io/dcpext0");
```
jannau, line 54: unrelated to the fix. this should use the display_config from display.c. I also don't see the need to handle this outside display.c. The called functions should just return immediately when running on device without dcp. I can fix that tonight

asdfugil:
get_device_info() call was moved to in `src/startup.c`, it's not "dropped".

PR #430, opened 2024-12-30
Update MacOS-native build instructions in README.md

DragonStar373:
Brew no longer includes lld with llvm, so now it has to be installed separately.

Not really _that_ important  since brew includes a notice about it after you install llvm, but either way that makes the current build instructions outdated.


marcan:
Please add the DCO signoff.

PR #432, opened 2025-01-20
Allow !no-map GPU reserved ranges and map them

asahilina:
This should make the behavior with and without u-boot (mostly) consistent. I think u-boot also adds mem ranges for `no-map`  reserved ranges (then the kernel doesn't map them as requested), but that shouldn't make much of a difference.

### DIFF
--- a/src/kboot.c
+++ b/src/kboot.c
@@ -37,6 +37,8 @@
 
 #define MAX_DISP_MAPPINGS 8
 
+#define MAX_MEM_REGIONS 32
+
 static void *dt = NULL;
 static int dt_bufsize = 0;
 static void *initrd_start = NULL;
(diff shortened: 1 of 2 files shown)


src/kboot.c
```diff
@@ -330,13 +332,72 @@ static int dt_set_memory(void)
...
-    u64 memreg[2] = {cpu_to_fdt64(dram_min), cpu_to_fdt64(dram_max - dram_min)};
+    struct {
+        fdt64_t start;
+        fdt64_t size;
+    } memreg[MAX_MEM_REGIONS] = {{cpu_to_fdt64(dram_min), cpu_to_fdt64(dram_max - dram_min)}};
+    int num_regions = 1;
+
+    int resv_node = fdt_path_offset(dt, "/reserved-memory");
+    if (resv_node < 0)
+        bail("FDT: '/reserved-memory' not found\n");
```
jannau, line 343: I think this should skip over the `fdt_for_each_subnode`. Our production devicetrees will always have `/reserved-memory` but for initial bringup it would be easier without `bail()`

src/kboot.c
```diff
@@ -330,13 +332,72 @@ static int dt_set_memory(void)
...
+
+    int node;
+    /* Find all reserved memory nodes */
+    fdt_for_each_subnode(node, dt, resv_node)
+    {
+        const char *name = fdt_get_name(dt, node, NULL);
+        const char *status = fdt_getprop(dt, node, "status", NULL);
+        if (status && !strcmp(status, "disabled"))
```
jannau, line 351: I'm surprised libfdt misses a function for this

jannau (commented):
Have you tested if u-boot still works with this?

asahilina:
> Have you tested if u-boot still works with this?

It seems to work fine. I think u-boot just merges everything together into its own memory map. The DT still contains a subset of memory ranges (main + the map ones this add), and u-boot still reports all reserved-memory ranges (including no-map) in the EFI mem map to the kernel.

PR #438, opened 2025-02-04
Update Dockerfile to Multi-Stage

multi-stager:
Hi,

Thanks for maintaining m1n1. I represent a research group investigating [multi-stage builds](https://docs.docker.com/build/building/multi-stage/). We recently refactored your Dockerfile to a multi-stage Dockerfile and found that it brings the following benefits: 

✅ Reduced final image size (from 1683MB to 75MB)  
✅ Minimized image vulnerabilities (verified via [Trivy](https://github.com/aquasecurity/trivy), from 3600 to 149)  

We hope this small improvement to the Dockerfile will be useful to you :)

Thanks,

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM debian:buster-slim
+### Build stage
+FROM debian:buster-slim AS builder
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && apt-get install -y build-essential bash curl git locales gcc-aarch64-linux-gnu libc6-dev-arm64-cross device-tree-compiler \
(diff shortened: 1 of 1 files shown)


jannau:
Splitting the Dockerfile is not appropriate. The only purpose of the Dockerfile is to provide a build environment. Note that the docker file does not even build the m1n1. Even if it did the build result would not be useful in docker as it is a boot loader which runs before Linux.

Please do not create automated pull request without investigating the purpose of the projects you fill PRs against.

multi-stager:
@jannau @WhatAmISupposedToPutHere This request was submitted by me manually (not spambot). We apologize for any problems caused by this request, our intention is just to help.

PR #439, opened 2025-02-08
A7-A11, T2 SoC cpufreq support

asdfugil:
This adds the tunables required for cpufreq to work properly on A7-A11, T2 SoCs, most prominently a register poke to enable voltage changes when the state is changed, as iBoot leaves the voltage controls disabled.

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -84,7 +84,7 @@ static int set_pstate(const struct cluster_t *cluster, uint32_t pstate)
                 return -1;
         }
         write64(cluster->base + CLUSTER_PSTATE, val);
-        if (poll32(cluster->base + CLUSTER_PSTATE, CLUSTER_PSTATE_BUSY, 0, CLUSTER_SWITCH_TIMEOUT) <
+        if (poll64(cluster->base + CLUSTER_PSTATE, CLUSTER_PSTATE_BUSY, 0, CLUSTER_SWITCH_TIMEOUT) <
             0) {
             printf("cpufreq: Timed out waiting for cluster %s P-State switch\n", cluster->name);
             return -1;
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -19,8 +19,13 @@
 #define CLUSTER_PSTATE_DESIRED2               GENMASK(15, 12)
 #define CLUSTER_PSTATE_APSC_BUSY              BIT(7)
 #define CLUSTER_PSTATE_DESIRED1               GENMASK(4, 0)
+#define CLUSTER_PSTATE_DESIRED1_S5L8960X      GENMASK(24, 22)
 
-#define CLUSTER_SWITCH_TIMEOUT 100
+#define PMGR_VOLTAGE_CTL_OFF_S5L8960X 0x20c00
+#define PMGR_VOLTAGE_CTL_OFF_T7000    0x23000
+#define PMGR_VOLTAGE_CTL_OFF_S8000    0xa0000
+
+#define CLUSTER_SWITCH_TIMEOUT 400
 
 struct cluster_t {
     const char *name;
@@ -42,6 +47,17 @@ struct feat_t {
 static u32 pstate_reg_to_pstate(u64 val)
 {
     switch (chip_id) {
+        case S5L8960X:
+        case T7000:
+        case T7001:
+            return FIELD_GET(CLUSTER_PSTATE_DESIRED1_S5L8960X, val);
+        case S8000:
+        case S8001:
+        case S8003:
+        case T8010:
+        case T8011:
+        case T8012:
+        case T8015:
         case T8103:
         case T6000:
         case T6001:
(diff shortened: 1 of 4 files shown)


asdfugil:
this is also a requirement for proper operation of the proxy with the physical uart, as the cpu is too slow otherwise

src/cpufreq.c
```diff
@@ -39,23 +44,41 @@ struct feat_t {
...
 static int set_pstate(const struct cluster_t *cluster, uint32_t pstate)
 {
     u64 val = read64(cluster->base + CLUSTER_PSTATE);
 
-    if (FIELD_GET(CLUSTER_PSTATE_DESIRED1, val) != pstate) {
-        val &= ~CLUSTER_PSTATE_DESIRED1;
-        val |= CLUSTER_PSTATE_SET | FIELD_PREP(CLUSTER_PSTATE_DESIRED1, pstate);
-        if (chip_id == T8103 || chip_id <= T6002) {
+    if (get_target_pstate(cluster) != pstate) {
+        if (chip_id == S5L8960X || chip_id == T7000 || chip_id == T7001) {
```
svenpeter42, line 62: I'd just use a switch/case with all chip_id listed explicitly here even if that results in some code duplication. Even the chip_id <= T6002 in the original code is a bit weird but now that we have three(?) different variants this is getting confusing.

src/cpufreq.c
```diff
@@ -89,11 +112,46 @@ int cpufreq_init_cluster(const struct cluster_t *cluster, const struct feat_t *f
     }
 
     /* Unknown */
-    write64(cluster->base + 0x440f8, 1);
+    if ((chip_id > T8015 && chip_id < S5L8960X) || chip_id < T7000)
```
svenpeter42, line 115: I'd really avoid all those > and < operations on chip_id and just hardcode a list where we need to do this write instead like e.g. the switch below

svenpeter42 (changes requested):

asdfugil:
Changes:
- Switch to `switch (chip_id)` for chip-specific operations, this removes all the comparison operations.

src/cpufreq.c
```diff
@@ -39,19 +44,44 @@ struct feat_t {
...
-    if (FIELD_GET(CLUSTER_PSTATE_DESIRED1, val) != pstate) {
-        val &= ~CLUSTER_PSTATE_DESIRED1;
-        val |= CLUSTER_PSTATE_SET | FIELD_PREP(CLUSTER_PSTATE_DESIRED1, pstate);
-        if (chip_id == T8103 || chip_id <= T6002) {
-            val &= ~CLUSTER_PSTATE_DESIRED2;
-            val |= CLUSTER_PSTATE_SET | FIELD_PREP(CLUSTER_PSTATE_DESIRED2, pstate);
+    if (pstate_reg_to_pstate(val) != pstate) {
+        switch (chip_id) {
+            case S5L8960X:
+            case T7000 ... T7001:
```
svenpeter42, line 67: This is now just hiding the comparisons that I'd like to avoid. Please get rid of the ... stuff as well and just open code all the chip_ids here.

src/cpufreq.c
```diff
@@ -39,19 +44,44 @@ struct feat_t {
...
+                val &= ~CLUSTER_PSTATE_DESIRED1_S5L8960X;
+                val |= CLUSTER_PSTATE_SET | FIELD_PREP(CLUSTER_PSTATE_DESIRED1_S5L8960X, pstate);
+                break;
+            case S8000 ... T8103:
+            case T6000 ... T6002:
+                val &= ~CLUSTER_PSTATE_DESIRED2;
+                val |= CLUSTER_PSTATE_SET | FIELD_PREP(CLUSTER_PSTATE_DESIRED2, pstate);
```
svenpeter42, line 74: please add a // fallthrough comment here

src/cpufreq.c
```diff
@@ -88,11 +118,44 @@ int cpufreq_init_cluster(const struct cluster_t *cluster, const struct feat_t *f
...
+    if (adt_get_reg(adt, pmgr_path, "reg", 0, &pmgr_reg, NULL) < 0) {
+        printf("Error getting /arm-io/pmgr regs\n");
+        return -1;
+    }
+
+    switch (chip_id) {
+        case S5L8960X:
+            write32(pmgr_reg + PMGR_VOLTAGE_CTL_OFF_S5L8960X, 1);
+            break;
+        case T7000 ... T7001:
```
svenpeter42, line 138: same as above: please don't rely on hidden comparisons here and just enumerate all chips

svenpeter42 (changes requested):

src/cpufreq.c
```diff
@@ -88,11 +118,44 @@ int cpufreq_init_cluster(const struct cluster_t *cluster, const struct feat_t *f
...
+        case S8000 ... T8015:
+            /*
+             * On T8015 this will result in the register being written
+             * two times (for two clusters). However, this is fine.
+             */
+            write32(pmgr_reg + PMGR_VOLTAGE_CTL_OFF_S8000, 1);
+            break;
+        case T8103 ... T8122:
+        case T6000 ... T6034:
+        default:
```
svenpeter42, line 150: Please also separate out the default case and just print that the chip is unsupported

asdfugil:
Changes:
- Remove *all* hidden comparison using `switch (chip_id)`, including the existing ones, now every supported chip is explicitly stated.
- Don't "support" T8122 and T6034
- Print chip is unsupported in default cases

PR #440, opened 2025-02-12
Fix Uncore PMC handling on T8015 and add more UPMC registers

asdfugil:
This PR fixes Uncore PMC handling on T8015 and add more UPMC registers information. This information is derived from xnu sources and testing on T8015 (which, among other things, have one uncore for both clusters and only have 8 counters). Correctness on M1 and above have not been tested.


svenpeter42:
happy to merge this if you rebase it onto the chickens cleanup and get it tested on m1/m2

asdfugil:
verified correctness on T6000 as well:

```
T8015 (global uncore):
msr(UPMESR0_EL1, 0x2) # event selection
msr(UPMECM0_EL1, 0x1) # event core mask
msr(UPMPCM_EL1, 1) # interrupt routing
msr(UPMC0_EL1, (1<<47)-1) # write counter
msr(UPMCR0_EL1, 0x1401) # enable counter

T8020 (per-cluster uncore):
msr(UPMESR0_EL1, 0x2) # event selection
msr(UPMECM0_EL1, 0x1) # event core mask
msr(UPMPCM_EL1, 1) # interrupt routing
msr(UPMC0_EL1, (1<<47)-1) # write counter
msr(UPMCR0_EL1, 0x140001) # enable counter
```

the counters are probably 64 (63) bit wide on t8112 though and I do not have that to test

src/chickens.c
```diff
@@ -91,6 +94,7 @@ const struct midr_part_features features_m2 = {
 // XXX figure out what features are actually available on M4
 const struct midr_part_features features_m4 = {
     .sleep_mode = SLEEP_NONE, // XXX probably new mode required
+    .uncore_version = UNCORE_V2,
```
svenpeter42, line 97: I'd probably just leave this out for now since we have no idea what M4 actually supports (and because M4 isn't really supported by any means for now)

src/exception.c
```diff
@@ -139,10 +139,21 @@ void exception_initialize(void)
...
     }
 
+    switch (cpu_features->uncore_version) {
+        case UNCORE_V1:
+            reg_clr(SYS_IMP_APL_UPMCR0, UPMCR0_IMODE_T8015);
+            break;
+        case UNCORE_V2:
+            reg_clr(SYS_IMP_APL_UPMCR0, UPMCR0_IMODE_T8020);
+            break;
+        default: // In boot CPU's EL3 cpu_features is not initialized yet but it is 0 on machines
```
svenpeter42, line 152: `case UNCORE_NONE` and then we should get a compiler warning if we ever add _V3 and forget to change it here

src/exception.c
```diff
@@ -409,19 +420,34 @@ void exc_fiq(u64 *regs)
...
+        uint64_t upmcr0_imode = 0;
+        switch (cpu_features->uncore_version) {
+            case UNCORE_V1:
+                upmcr0_imode = UPMCR0_IMODE_T8015;
+                break;
+            case UNCORE_V2:
+                upmcr0_imode = UPMCR0_IMODE_T8020;
+                break;
+            default:
+                panic("Unhandled UPMC IRQ\n"); // FIQ storm is coming
```
svenpeter42, line 439: uh, how? I don't see how the default case can be reached. `UNCORE_NONE` would bail out earlier in the outer if

svenpeter42 (changes requested):
looks good except for the small comments. please also rebase on main to make the CI failures disappear

src/exception.c
asdfugil: this should be unreachable as long as there's no _V3 and similar

src/exception.c
```diff
@@ -409,19 +420,33 @@ void exc_fiq(u64 *regs)
...
+    if (cpu_features->uncore_version) {
+        uint64_t upmcr0_imode = 0;
+        switch (cpu_features->uncore_version) {
+            case UNCORE_V1:
+                upmcr0_imode = UPMCR0_IMODE_T8015;
+                break;
+            case UNCORE_V2:
+                upmcr0_imode = UPMCR0_IMODE_T8020;
+                break;
+            default:
```
svenpeter42, line 438:
```suggestion
            case UNCORE_NONE:
```

svenpeter42 (changes requested):

svenpeter42 (approved):

PR #445, opened 2025-02-19
GitHub Actions: Build on an ARM processor

cclauss:
https://github.com/actions/runner-images?tab=readme-ov-file#available-images

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-2024.04-arm
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   format-check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-2024.04-arm
 
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it


svenpeter42:
So this doesn't work because there is no -arm image as far as I can tell. Also doesn't really matter what the host is since it'll be compiled for arm either way.

cclauss:
> doesn't really matter what the host is

The job will run in half the time. ;-)

svenpeter42:
> > there is no -arm image
> 
> I do not understand that.

I approved the job and it never started because the -arm variant doesn't exist. It's also nowhere to be found in the documentation link you provided.

cclauss:
They still seem to be there... https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

svenpeter42:
"The arm64 Linux runner is in public preview and subject to change."

And this also just didn't work, see https://github.com/AsahiLinux/m1n1/actions/runs/13423890788/job/38989485442 "The job has exceeded the maximum execution time while awaiting a runner for 24h0m0s"

mkurz:
The reason why this is not working is because the image `ubuntu-2024.04-arm` does not exist. The correct name is `ubuntu-24.04-arm`. See https://github.com/actions/partner-runner-images/?tab=readme-ov-file#available-images

> They still seem to be there... https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Actually also here the image is called `ubuntu-24.04-arm`...

.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-2024.04-arm
```
mkurz, line 18:
```suggestion
    runs-on: ubuntu-24.04-arm
```

.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 jobs:
   format-check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-2024.04-arm
```
mkurz, line 18:
```suggestion
    runs-on: ubuntu-24.04-arm
```

mkurz (commented):
Fixed the image names.

svenpeter42:
Still not gonna accept this because "The arm64 Linux runner is in public preview and subject to change." like I mentioned above

PR #426, opened 2024-12-13
m1n1 :  add initial support for T6030

a-ramses:
The system counter frequency (CNTFRQ_EL0) varies between the efficiency and performance cores on the M3. The efficiency cores run at 1 GHz, as required by Armv8.6-A, while the performance cores continue to utilize Apple's standard frequency of 24 MHz. Hypervisor still does not work yet due to missing counter scaling bit in order to switch in EL1.

With that said Linux boots on the M3 Pro using all 11 cores(performance and efficiency) with a very limited configuration:
interrupt controller and uart for now able to reach penguins. 
Linux DT : https://github.com/a-ramses/linux/tree/t6030-j514s.
This patch relies on changes to parse "spmi-cha**" @jannau

PR #450, opened 2025-03-17
T603x bringup

step21:
Re-submission of https://github.com/AsahiLinux/m1n1/pull/426 with squashed format commits, socs in numerical order and placing new define macros in midr.h instead of in chickens.


step21:
thx!

PR #457, opened 2025-03-29
RFC: start rusting the ADT code

chadmed:
Don't really see a point to merging this until it's done. This is just to track the work and make sure I'm on the right track.


Makefile
```diff
@@ -67,10 +67,9 @@ endif
 # Required for no_std + alloc for now
 export RUSTC_BOOTSTRAP=1
 RUST_LIB := librust.a
-RUST_LIBS :=
+RUST_LIBS += $(RUST_LIB)
```
svenpeter42, line 70: You can drop RUST_LIBS entirely by using RUST_LIB inside OBJECTS directly

chadmed:
Rebased on top of main, and refactored API to make a bit more sense on the Rust side.

`time proxyclient/tools/chainload.py build/m1n1.macho` 
on main: 2.171s
on riir/adt: 2.144s

rust/src/adt.rs
```diff
@@ -0,0 +1,846 @@
...
+    BadOffset = -4,
+    BadPath = -5,
+    BadNCells = -14,
+    BadValue = -15,
+    BadLength = -20,
+}
+
+const ADT_ALIGN: i32 = 4;
+
+#[repr(C)]
```
WhatAmISupposedToPutHere, line 19: Fat pointers like &[T] do not have a stable C representation despite rust allowing you to put #[repr(C)] on a struct containing one.

rust/src/adt.rs
```diff
@@ -0,0 +1,846 @@
...
+pub struct ADT<'a> {
```
WhatAmISupposedToPutHere, line 20: Is it even constructed anywhere? looks like a dead struct.

rust/src/adt.rs
```diff
@@ -0,0 +1,846 @@
...
+
+        if b[..len].contains('@') && a.as_bytes().get(len) == Some(&b'@') {
+            return true;
+        }
+
+        false
+    }
+
+    /// Trace a given ADT path and return the offset of the final node.
+    pub fn path_offset_trace(path: &str, mut offsets: Option<&mut [i32]>) -> Result<i32, AdtError> {
```
WhatAmISupposedToPutHere, line 75: Here and everywhere else where you use node offsets - unsound api from Rust's POV as the offsets are invalidated when you change the ADT, but i am not sure how to properly model it for now and whether it is worth it.

rust/src/adt.rs
chadmed: Better to declare it as `*const u8` then? The struct doesn't actually do anything, it was a leftover from a previous architecture. Probably fine to just get rid of it and move the functions in the ADT impl out of it.

chadmed:
I've changed the internal API to use absolute pointers to the nodes/properties instead of offsets. The C bindings calculate these relative to the head of the ADT, then derives offsets to give back to C code.

The public Rust-side API deals with neither pointers nor offsets. Instead, tracing a path will yield a slice of static references where required, which can be indexed into/iterated over/whatever. Internally we still need to deal with raw pointers since all we get from iBoot is a pointer to the head of the ADT, however this can be hidden from consumers now.

rust/src/adt.rs
```diff
@@ -0,0 +1,909 @@
...
+        Err(AdtError::BadOffset)
+    } else {
+        Ok(())
+    }
+}
+
+/// Determine if an ADT node's name is equal to some string bounded by a length.
+/// This is required to match the semantics expected when doing a full ADT path
+/// trace.
+fn node_names_equal(a: &str, b: &str, len: usize) -> bool {
```
WhatAmISupposedToPutHere, line 56: pre-slice b in the caller?

rust/src/adt.rs
```diff
@@ -0,0 +1,909 @@
...
+/// into this array.
+pub fn get_reg_container(
+    nodes: &[Option<&'static ADTNode>],
+    pname: &str,
+    i: i32,
+) -> Result<(u64, u64), AdtError> {
+    let head = unsafe { ADTNode::from_ptr(adt as *const ADTNode)? };
+
+    if nodes[0].is_some() {
+        if nodes[0].unwrap().as_ptr() == unsafe { adt as *const u8 } {
```
WhatAmISupposedToPutHere, line 91: if let Some(...) = nodes[0]

rust/src/adt.rs
```diff
@@ -0,0 +1,909 @@
...
+            return Err(AdtError::BadOffset);
+        }
+    }
+    let mut cursor: usize = nodes.len() - 1;
+
+    while nodes[cursor].is_none() {
```
WhatAmISupposedToPutHere, line 97: panics if nodes does not contain at least one Some. Can that happen?

rust/src/adt.rs
```diff
@@ -0,0 +1,909 @@
...
+            )
+        }
+    }
+
+    /// Search the node for a property with the given name.
+    pub fn named_prop(&self, name: &str) -> Result<&'static ADTProperty, AdtError> {
+        let p_cnt = self.property_count;
+        let mut p = self.first_property()?;
+
+        for _i in 0..p_cnt {
```
WhatAmISupposedToPutHere, line 347: `s/_i/_/` 

same below

rust/src/adt.rs
```diff
@@ -0,0 +1,909 @@
...
+            p = p.next_property_mut()?;
+        }
+        Err(AdtError::NotFound)
+    }
+
+    pub fn next_sibling(&self) -> Result<&'static ADTNode, AdtError> {
+        let mut cnt = self.child_count;
+        let mut c = self.first_child()?;
+
+        while cnt > 0 {
```
WhatAmISupposedToPutHere, line 378: for instead of while?

WhatAmISupposedToPutHere (commented):
lgtm in general, mostly some minor style nitpicks

rust/src/adt.rs
chadmed: Got rid of the need to slice the strings at all

rust/src/adt.rs
chadmed: Added sanity check made it safe to just unwrap nodes[0] unconditionally

rust/src/adt.rs
chadmed: Added a sanity check to make sure we have at least one Some

WhatAmISupposedToPutHere (approved):

svenpeter42 (approved):

PR #458, opened 2025-04-07
utils/proxy: Fix cpu_sleep global sleep handling

svenpeter42:
da3519de142 introduced the additional argument cpufeat_global_sleep to cpu_sleep but did neither update its prototype nor the call inside proxy.c. Instead of adding a new argument just directly load the value of cpufeat_global_sleep and use it instead.

Fixes: da3519de142


asdfugil (approved):

jannau (approved):
I think you can only request reviews from project members.

The ppc build error can be safely ignored as there are known issues with ppc copr builds, not that it would be an important target anyway.

PR #460, opened 2025-04-13
cleanup chickens.c

svenpeter42:
- [x] cpu_sleep looks suspicious, double check it's correct -> it still looks suspicious but produces similar assembly code like the old version. someone should double check though
- [x] test this as stage 1 

Only tested on t8103 so far


src/chickens.c
```diff
@@ -180,13 +185,12 @@ const char *init_cpu(void)
         sysop("isb");
     }
 
-    if (part >= MIDR_PART_T8101_ICESTORM && part != MIDR_PART_T8301_THUNDER) {
```
jannau: the special case is T8301 and this drops support for the unsupported T8030

src/chickens.c
```diff
@@ -32,10 +32,12 @@ void init_t6031_sawtooth(int rev);
 void init_t6031_everest(int rev);
 
 bool cpufeat_actlr_el2, cpufeat_fast_ipi, cpufeat_mmu_sprr;
-bool cpufeat_global_sleep, cpufeat_workaround_cyclone_cache;
+bool cpufeat_workaround_cyclone_cache;
+enum cpufeat_sleep_mode cpufeat_sleep_mode;
 
 struct midr_part_features {
     bool workaround_cyclone_cache;
+    enum cpufeat_sleep_mode sleep_mode;
```
jannau: please put the `enum` first to avoid unnecessary padding

jannau (commented):
Nice, except for the field order in `midr_part_features` no chnges needed.

src/chickens.c
```diff
@@ -164,9 +177,6 @@ const char *init_cpu(void)
         sysop("isb");
 
         msr(SYS_IMP_APL_AMX_CTX_EL1, core);
-
-        /* T8030 SPRR is different */
```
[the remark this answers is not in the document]
svenpeter42: huh, true. i'll fix the commit description while i'm rebasing anyway.

svenpeter42:
I still think the sleep function looks very suspicious, especially isb -> wfi in a loop. Feels like there's something missing there  even in the assembly version that would actually require that isb.

Would be great to test this on some !t8103 machines. I don't think it breaks anything but it's also not urgent that we merge this.

jannau:
I ignored the loop. It's intentionally sketchy because the sleep is eternal. CPU wake up was left as TODO. I guess the `isb` could be moved out of the loop.

jannau (approved):

PR #462, opened 2025-04-23
Fix PAN issue

flatz:
It seems like that check if PAN is supported (introduced in [869d2ae35c31e14935a63492070699aa359df717](https://github.com/AsahiLinux/m1n1/commit/869d2ae35c31e14935a63492070699aa359df717)) by reading MSR _ID_AA64MMFR1_EL1_ may fail under some circumstances which happens everytime at least on my Mac Mini 2023 (T8112) that cause early XNU's panic with this exception in m1n1's log:

```
TTY> Running in EL2
TTY> MPIDR: 0x80000001
TTY> Registers: (@0x804ad7dd0)
TTY>   x0-x3: 0000000000000001 0000000000000000 0000000000000000 0000000000000000
TTY>   x4-x7: 0000000000000000 0000000000000000 0000000000000000 0000000000000000
TTY>  x8-x11: 0000000030901085 0000000000000001 a005a500f005f00f f005a500f00ff00f
TTY> x12-x15: 0000000000000000 0000000000000000 0000000000000000 0000000000000000
TTY> x16-x19: 0000000000000000 0000000000000000 00000000fffffff5 0000000803541260
TTY> x20-x23: 000000080353800a 0000000803541000 0000000000000001 0000000000000000
TTY> x24-x27: 0000000000000000 0000000000000000 0000000000000000 0000000000000000
TTY> x28-x30: 0000000000000000 00000008034a8f04 0000000002000000
TTY> PC:       0x2000000 (rel: 0xfffffff7feb80000)
TTY> SP:       0x804ad7ed0
TTY> SPSR:     0x10c9
TTY> FAR:      0x2000000
TTY> ESR:      0x86000006 (instruction abort (current))
TTY> L2C_ERR_STS: 0x0
TTY> L2C_ERR_ADR: 0x0
TTY> L2C_ERR_INF: 0x0
TTY> E_LSU_ERR_STS: 0x0
TTY> E_FED_ERR_STS: 0x0
TTY> E_MMU_ERR_STS: 0x0
```

Fixing this by caching PAN bit value at early m1n1's boot.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -542,7 +542,7 @@ void mmu_init(void)
         mmu_init_sprr();
 
     // Enable EL0 memory access by EL1
-    if (supports_pan())
+    if (has_pan)
         msr(PAN, 0);
 
     // RES1 bits
@@ -562,7 +562,7 @@ static void mmu_secondary_setup(void)
         mmu_init_sprr();
 
     // Enable EL0 memory access by EL1
-    if (supports_pan())
+    if (has_pan)
         msr(PAN, 0);
 
     // RES1 bits
--- a/src/start.S
+++ b/src/start.S
@@ -81,6 +81,18 @@ _start:
     mov w0, 'n'
     bl debug_putc
 
+    adrp x1, _bss_start
+    add x1, x1, :lo12:_bss_start
+    adrp x2, _bss_end
+    add x2, x2, :lo12:_bss_end
+    cmp x2, x1
+    beq 1f
+1:
+    str xzr, [x1], #8
+    cmp x1, x2
+    b.lo 1b
+2:
+
     adrp x0, _base
     mov x20, x0
     adrp x1, _rela_start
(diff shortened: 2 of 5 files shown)


svenpeter42:
So I've run into this problem before but I'm not convinced `supports_pan()` is the actual root cause:
*Something* goes wrong while the CPU of a secondary core is in `mmu_secondary_setup` (which then corrupts LR I think) and if you change the C code in there just enough to cause a slightly different assembly (probably removing an additional function call which stores LR on the stack) the problem you're seeing goes away. I haven't been able to figure out what exactly causes it though.

svenpeter42:
Even switching from clang to gcc mitigates this for me fwiw.
I don't think it makes sense to merge this because it's a workaround and the issue might just appear again when something else is changed. I'd really like to find the root cause instead but haven't been very successful so far. Bisecting this has been almost useless. I haven't ever run into this with v1.4.16 but that could also just be a coincidence. Maybe we can just create an issue for now and write down what we've figured out so far.

PR #466, opened 2025-05-11
Move creating gpu blobs to m1n1.

WhatAmISupposedToPutHere:
Instead of doing gpu initialization partially in m1n1 and partially in kernel, build all the blobs in m1n1 and put them in the device tree to be picked up by the kernel. Most of the code here is copied and pasted kernel code, with some fixups


svenpeter42:
looks like this also breaks the CI because https://src.fedoraproject.org/rpms/m1n1/raw/main/f/m1n1-rust-deps.patch doesn't apply anymore. no idea where that file comes from though or why we need it.

@davide125 @Conan-Kudo

davide125:
It's here: https://src.fedoraproject.org/rpms/m1n1/blob/rawhide/f/m1n1-rust-deps.patch . We need it because in Fedora we build against the system packages, not the vendored ones.

svenpeter42:
`<kettenis> regarding m1n1 and rust on OpenBSD; yes that is a huge problem since you can't easily build your own standard library for a cross compiler`

let's not break m1n1 stage2 builds for all openbsd users. so we either need to convert this all to C (ugh...) or allow builds without rust and gpu support and make sure we don't shot ourselves in the foot with that

src/kboot.c
```diff
@@ -47,7 +47,9 @@ static char *chosen_params[MAX_CHOSEN_PARAMS][2];
 
 extern const char *const m1n1_version;
 
+#ifdef DO_GPU_INIT
 int dt_set_gpu(void *dt);
```
svenpeter42, line 51: I'd probably do something like 
```suggestion
#ifdef DP_GPU_INIT
int dt_set_gpu(void *dt);
#else
static int dt_set_gpu(void *dt)
{
    printf("Cannot setup GPU data in devicetree because m1n1 was built without rust support");
}
#endif
```
here just to be as verbose as possible

svenpeter42 (commented):
Lina's sign-off is missing, my other comment is optional and can be ignored. I also didn't really review the rust code because it's been copied from the kernel and we know that it works

WhatAmISupposedToPutHere:
> Lina's sign-off is missing, my other comment is optional and can be ignored. I also didn't really review the rust code because it's been copied from the kernel and we know that it works

It is not a 1 to 1 copy, but i have a special kernel build that compares the kernel blob with m1n1 one, and those matched on t6000

svenpeter42 (approved):

svenpeter42:
Let's see if jannau wants to review all this magic since he's more familiar with both rust and the gpu driver but I'm fine with merging it

src/kboot_gpu.c
```diff
@@ -674,5 +687,54 @@ int dt_set_gpu(void *dt)
...
+
+    int adt_sgx_path[8];
+    if (adt_path_offset_trace(adt, "/arm-io/sgx", adt_sgx_path) < 0)
+        bail("ADT: GPU: Failed to get sgx\n");
+
+    if (adt_get_reg(adt, adt_sgx_path, "reg", 0, &gpu_base, NULL) < 0)
+        bail("ADT: GPU: Failed to get sgx reg 0\n");
+
+    u32 base_pstate;
+    if (ADT_GETPROP(adt, sgx, "gpu-perf-base-pstate", &base_pstate) < 0)
```
jannau, line 712: this property does not exists on all devices. It is at least missing on M1 mac mini and imac

src/kboot_gpu.c
```diff
@@ -674,5 +687,54 @@ int dt_set_gpu(void *dt)
...
+    if (has_cs_afr) {
+        n_perf_states_cs = perf_states_cs->count;
+        n_perf_states_afr = perf_states_afr->count;
+        pstate_cs_raw = perf_states_cs->states;
+        pstate_afr_raw = perf_states_afr->states;
+    }
+    if (rust_fill_gpu_initdata(
+            perf_state_table_count, perf_state_count, perf_states, max_pwr, core_leak, sram_leak,
+            cs_leak, afr_leak, n_perf_states_cs, pstate_cs_raw, n_perf_states_afr, pstate_afr_raw,
+            adt_is_compatible(adt, 0, "J180dAP"), base_pstate, uat_ttb_base, num_cores,
```
jannau, line 727: handling `J180d` specially should not be necessary. This seems to be an bug in our devicetrees. The J475d ADT uses the exact same properties as specified in `t6022-j180d.dts`

src/kboot_gpu.c
```diff
@@ -37,6 +38,18 @@ struct aux_perf_states {
     struct aux_perf_state states[];
 };
 
+int32_t rust_gpu_initdata_size(uint32_t compat_maj, uint32_t compat_min, size_t *data_a,
+                               size_t *data_b, size_t *globals);
+
+int32_t rust_fill_gpu_initdata(size_t perf_state_table_count, size_t perf_state_count,
```
jannau, line 44: can we use a struct for this?

src/kboot_gpu.c
```diff
@@ -37,6 +38,18 @@ struct aux_perf_states {
...
+                               const struct perf_state *c_perf_states, uint32_t *max_pwr,
+                               float *core_leak, float *sram_leak, float *cs_leak, float *afr_leak,
+                               size_t n_perf_states_cs, const struct aux_perf_state *pstates_cs,
+                               size_t n_perf_states_afr, const struct aux_perf_state *pstates_afr,
+                               uint32_t is_j180, uint32_t base_pstate, uint64_t uat_ttb_base,
+                               uint32_t num_cores, uint32_t compat_maj, uint32_t compat_min,
+                               uint32_t gpu_rev_id, void *data_a, void *data_b, void *globals);
```
jannau, line 51: I'd feel more comfortable if we pass the size `data_a`, `data_b` and `globals` as well.

src/kboot.c
```diff
@@ -2569,7 +2579,9 @@ int kboot_prepare_dt(void *fdt)
     dt_bufsize = fdt_totalsize(fdt);
     assert(dt_bufsize);
 
-    dt_bufsize += 6 * SZ_16K; // Add 96K of buffer for modifications
+    // Add 96K of buffer for modifications
+    // Thats not enough for gpu data alone, do 288
```
jannau, line 2583: this comment reads strange now. How about
```
    // Add 96K of buffer for modifications and 192K for GPU init data
```

jannau (commented):
First review of the C code

src/kboot_gpu.c
WhatAmISupposedToPutHere: The caller is supposed to allocate exactly as much as `rust_gpu_initdata_size` asked for, so not sure what value that adds?

src/kboot_gpu.c
WhatAmISupposedToPutHere: We can, i just find structs to be a bit inconvenient to keep in sync with rust/c

jannau: but keeping 22 function arguments in sync is convenient? I find the the function argument variant harder to validate and would assume it's more likely to attract bugs. We could use bindgen.

WhatAmISupposedToPutHere: no, they are both terrible tbh. i'll look into bindgen

PR #471, opened 2025-05-30
Makefile: Optionally build rust code with -Z build-std

kettenis:
Allows building m1n1's rust components on systems without rustup support.  Removes the need to install the
aarch64-unknown-node-softfloat toolchain.  But it does require the rust-src component to be installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,12 @@ CFG += CHAINLOADING
 RUST_LIBS += $(RUST_LIB)
 endif
 
+ifeq ($(BUILDSTD),1)
+CARGO_FLAGS := -Z build-std=alloc,core
+else
+CARGO_FLAGS :=
+endif
+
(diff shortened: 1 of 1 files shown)


kettenis:
Alternatively we could make this the default. But that probably needs changes in the CI scripts and distro packaging.

svenpeter42 (approved):
looks good to me, someone else should comment on if making this the default makes sense since i haven't ever looked into distro packaging.

jannau (approved):
I was going to suggest documenting this option but I discovered that we don't document the rust build at all.

jannau:
> looks good to me, someone else should comment on if making this the default makes sense since i haven't ever looked into distro packaging.

Fedora has `rust-std-static-aarch64-unknown-none-softfloat` packaged (other distros probably as well) and it exists iirc also as rustup toolchain for linux so not recompiling rust standard library by default is useful.

PR #474, opened 2025-06-28
rust: fix the litany of compiler warnings

chadmed:
When built from the main tree, rustc emits myriad warnings about various things. These are mostly a result of the vendored crates becoming stale.

We cannot unvendor fatfs due to a reliance on features present in the upstream tree that have not been released yet, but the most recent revision now builds fine with released versions of its dependencies. Let's update our vendored version of fatfs, and get rid of all these stale vendored crates to clean up rustc's output.

While we're at it, since `new_uninit` is stable as of rust-1.82, let's guard that feature flag behind a version check. At some point in the future, we should remove all of these checks and simply make 1.82 the MSRV.

Tested as m1n1 stage1, no regressions in chainloading noticed.


chadmed:
Build failures can be safely ignored. It is failing to apply a downstream patch to fatfs that is now incorporated upstream.

svenpeter42 (approved):
LGTM, i'd merge this after #477 because I really don't like those annoying false CI failures

chadmed:
Rebased to get rid of the packit CI failures

PR #477, opened 2025-08-03
[RFC] Drop PackIt integration

jannau:
PackIt ends up more annoying than helpful since source code and rpm spec file are not in sync. Drop it since importing the spec file from fedora and keeping it in sync is not an attractive option either.


jannau:
See the false CI failures in #457 and #474

PR #479, opened 2025-08-04
Only compile assembly codes as armv8.2-a

asdfugil:
Compile C code as base armv8-a, and compile assembly codes as armv8.2-a to allow m1n1 to be properly compiled for older SoCs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,10 +54,10 @@ BASE_CFLAGS := -O2 -Wall -g -Wundef -Werror=strict-prototypes -fno-common -fno-P
 	-Wsign-compare -Wunused-parameter -Wno-multichar \
 	-ffreestanding -fpic -ffunction-sections -fdata-sections \
 	-nostdinc -isystem $(shell $(CC) -print-file-name=include) -isystem sysinc \
-	-fno-stack-protector -mstrict-align -march=armv8.2-a \
+	-fno-stack-protector -mstrict-align -Wa,-march=armv8.2-a \
 	$(EXTRA_CFLAGS)
 
-CFLAGS := $(BASE_CFLAGS) -mgeneral-regs-only
+CFLAGS := $(BASE_CFLAGS) -mgeneral-regs-only -march=armv8-a -mno-outline-atomics
 
 CFG :=
 ifeq ($(RELEASE),1)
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -49,12 +49,14 @@ QUIET :=
 endif
 endif
 
+ASM_ARCH := armv8.2-a
```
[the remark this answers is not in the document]
asdfugil: The assembler arguments (-Wa,-march) are still needed to allow armv8.2 instructions in the .S files, so this variable tries to make it so there are only one architecture version in the source tree

asdfugil:
updated to specify armv8.2-a directly whenever needed

asdfugil:
the build failures appears to be because of the build environment not supplying some runtime builtins that are required when LSE atomics are unavailable

src/utils.h
```diff
@@ -266,15 +266,15 @@ static inline void write64_lo_hi(u64 addr, u64 val)
 #define __mrs(reg)                                                                                 \
     ({                                                                                             \
         u64 val;                                                                                   \
-        __asm__ volatile("mrs\t%0, " #reg : "=r"(val));                                            \
+        __asm__ volatile(".arch armv8.2-a\nmrs\t%0, " #reg : "=r"(val));                           \
```
jannau, line 269: please break the asm statement into 2 lines like
```
        __asm__ volatile(".arch armv8.2-a\n"                                                       \
                         "mrs\t%0, " #reg : "=r"(val));                                            \
```

Makefile
```diff
@@ -54,7 +54,7 @@ BASE_CFLAGS := -O2 -Wall -g -Wundef -Werror=strict-prototypes -fno-common -fno-P
 	-Wsign-compare -Wunused-parameter -Wno-multichar \
 	-ffreestanding -fpic -ffunction-sections -fdata-sections \
 	-nostdinc -isystem $(shell $(CC) -print-file-name=include) -isystem sysinc \
-	-fno-stack-protector -mstrict-align -march=armv8.2-a \
+	-fno-stack-protector -mstrict-align -Wa,-march=armv8.2-a \
```
jannau, line 57: I don't think this is enough for armv8-a under all circumstances. I think you have to explicitly set `-march=armv8-a` and add `-mno-outline-atomics` to make sure we end up with base armv8-a

jannau (commented):
I'm not sure if I'm happy with this change. I would prefer if the HV continues to use LSE atomics if available.
Could you live with enabling this only for old SoCs via a config variable?

asdfugil:
> spinlock (and hv) still use built-in atomics and with `-march=armv8.2-a` those translate to LSE atomics

yeah but the asm block used to use LSE atomics as well but this was changed so m1n1 could work on armv8.0 https://github.com/AsahiLinux/m1n1/commit/869d2ae35c31e14935a63492070699aa359df717

asdfugil:
i needed hypervisor codes after all so I should close this PR and see what to do later

PR #486, opened 2025-08-13
kboot_gpu: Fix booting with v6.17-rc1 upstream DT

jannau:
Booting with Linux 6.17-rc1 device trees fails due to missing "operating-points-v2" properties. Check for the downstream only "apple,agx-t*" compatible strings and disable the gpu node if those are not found.
This serves as temporary fix until
https://github.com/AsahiLinux/m1n1/pull/466 is merged.

### DIFF
--- a/src/kboot_gpu.c
+++ b/src/kboot_gpu.c
@@ -500,6 +500,23 @@ int dt_set_gpu(void *dt)
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


svenpeter42 (approved):
ouch. do we have to create a bug fix release with just this commit?

PR #489, opened 2025-08-27
kboot_atc: Add fuses for t6000, t6002, and t8112

svenpeter42:
Untested since I have none of that hardware, I've only ran the script and copy/pasted the results.


jannau (commented):
Code looks reasonable and is tested on t6001 and t8112 using commit 90e2c56eccad of the atcphy/dwc3 upstream submission branch.

jannau (approved):

PR #508, opened 2025-10-20
usb: skip usb_i2c_restore_irqs on models where i2c is not present

yuyuyureka:
This prevents a crash while chainloading

### DIFF
--- a/src/usb.c
+++ b/src/usb.c
@@ -315,7 +315,7 @@ void usb_init(void)
         return;
 
     /*
-     * M3 models do not use i2c, but instead SPMI with a new controller.
+     * M3/M4 models do not use i2c, but instead SPMI with a new controller.
      * We can get USB going for now by just bringing up the phys.
      */
     if (adt_path_offset(adt, "/arm-io/nub-spmi-a0/hpm0") > 0) {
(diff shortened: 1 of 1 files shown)


src/usb.c
```diff
@@ -399,9 +401,13 @@ void usb_i2c_restore_irqs(const char *i2c_path, bool force)
 
 void usb_hpm_restore_irqs(bool force)
 {
-    if (adt_is_compatible(adt, 0, "J180dAP"))
+    if (adt_is_compatible(adt, 0, "J180dAP")) {
         usb_i2c_restore_irqs("/arm-io/i2c3", force);
-    usb_i2c_restore_irqs("/arm-io/i2c0", force);
+    } else if (adt_is_compatible(adt, 0, "J773gAP")) {
```
jannau: This needs to be a plain if. The Mac Pro ("J180dAP") has hpm devices on i2c0 and i2c3

src/usb.c
```diff
@@ -335,6 +335,8 @@ void usb_init(void)
 
     if (adt_is_compatible(adt, 0, "J180dAP") && usb_init_i2c("/arm-io/i2c3") < 0)
         return;
+    if (adt_is_compatible(adt, 0, "J773gAP") && usb_init_i2c("/arm-io/i2c2") < 0)
```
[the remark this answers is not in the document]
yuyuyureka: Not knowing more about how USB C works in these devices, I saw `/arm-io/i2c2/atcrt{0,1,3}` in the ADT and assumed this meant i2c2 would be the correct device node (atc = apple type-c?). It is completely possible that this is garbage.

I do see a /nub-spmi-a0/hpm0 too.

Either way, the usb_i2c_restore_irqs on /arm-io/i2c0 must not run on the M4 Mac Mini, because it crashes when checking for usbc,manager in the adt.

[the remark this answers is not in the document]
svenpeter42: HPM is the type-c controller that used to be on i2c. atcrt sounds like they possibly added a retimer chip which I'm going to hate because it appears to be stateful :(

src/usb.c
```diff
@@ -399,6 +399,19 @@ void usb_i2c_restore_irqs(const char *i2c_path, bool force)
 
 void usb_hpm_restore_irqs(bool force)
 {
+    /*
+     * Do not try to restore irqs on M3/M4 which don't use i2c
+     */
+    if (adt_path_offset(adt, "/arm-io/nub-spmi-a0/hpm0") > 0)
+        return;
```
jannau: This starts growing to a level where we probably should start thinking of adding some sort of device abstraction like Sven did with CPU initialization and I started for display initialization.

jannau (approved):
Comments are random thoughts while looking at the change, no need to change anything.

src/usb.c
jannau: that would be a possibility as well

PR #516, opened 2025-11-30
usb: fix typo in #488

mildsunrise:
silly typo that AFAICT currently makes us ignore all HPMs


mildsunrise:
ah, i didn't fix the second instance of the typo... will do tomorrow

PR #517, opened 2025-12-02
SPMI driver

mildsunrise:
expands the Python driver / regmaps with many of the things in https://github.com/AsahiLinux/docs/pull/225 and implements the basic functionality as a C driver

### DIFF
--- a/src/spmi.c
+++ b/src/spmi.c
@@ -1,0 +1,208 @@
+/* SPDX-License-Identifier: MIT */
+
+#include "spmi.h"
+#include "adt.h"
+#include "malloc.h"
+
+#define SPMI_STATUS 0x00
+#define SPMI_CMD    0x04
+#define SPMI_REPLY  0x08
+
+#define SPMI_CMD_EXTRA  GENMASK(31, 16)
+#define SPMI_CMD_ACTIVE BIT(15)
+#define SPMI_CMD_ADDR   GENMASK(14, 8)
+#define SPMI_CMD_OPCODE GENMASK(7, 0)
+
+#define SPMI_REPLY_FRAME_PARITY GENMASK(31, 16)
+#define SPMI_REPLY_ACK          BIT(15)
+#define SPMI_REPLY_ADDR         GENMASK(14, 8)
+#define SPMI_REPLY_OPCODE       GENMASK(7, 0)
+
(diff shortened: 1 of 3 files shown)


src/spmi.c
```diff
@@ -0,0 +1,210 @@
...
+            printf("spmi: Reply was shorter than expected\n");
+            return -1;
+        }
+        u32 data = read32(dev->base + SPMI_REPLY);
+        for (size_t j = 0; (j < 4) && (i < len_out);)
+            data_out[i++] = data >> (j++ * 8);
+    }
+
+    if (FIELD_GET(SPMI_REPLY_FRAME_PARITY, reply) != MASK(len_out))
+        return -2;
```
svenpeter42: I would either always return -1 or add defines for the various errors

src/spmi.c
```diff
@@ -0,0 +1,210 @@
...
+        printf("spmi: Invalid slave address %u\n", addr);
+        return -3;
+    }
+    if (len_out > 16) {
+        printf("spmi: Invalid out size %lu\n", len_out);
+        return -3;
+    }
+
+    // ensure FIFOs are in the correct state
+
```
svenpeter42: nit: I'd drop the empty line here (and below after the next comment)

mildsunrise: oki! wasn't sure if you were also including the next next comment (for consistency), i removed the empty line there too but lmk if wrong

src/spmi.c
```diff
@@ -0,0 +1,210 @@
...
+#define SPMI_REPLY_ACK          BIT(15)
+#define SPMI_REPLY_ADDR         GENMASK(14, 8)
+#define SPMI_REPLY_CMD          GENMASK(7, 0)
+
+#define SPMI_STATUS_RX_EMPTY BIT(24)
+#define SPMI_STATUS_RX_COUNT GENMASK(23, 16)
+#define SPMI_STATUS_TX_EMPTY BIT(8)
+#define SPMI_STATUS_TX_COUNT GENMASK(7, 0)
+
+#define SPMI_CMD__RESET    0x10
```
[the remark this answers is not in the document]
mildsunrise: yeah, same... I think I'll change it to `SPMI_OPCODE_` (or `SPMI_OPC_`) which is what Linux uses

PR #518, opened 2025-12-07
T8122 bringup

jannau:
Adds chickens, midr and bringup UART address and refactors dapf parsing

### DIFF
--- a/src/chickens_everest.c
+++ b/src/chickens_everest.c
@@ -36,6 +36,43 @@ static void init_common_everest(void)
     reg_clr(s3_0_c15_c2_4, BIT(0) | BIT(1) | BIT(16) | BIT(17) | BIT(18) | BIT(22));
 }
 
+void init_t8122_everest(int rev)
+{
+    UNUSED(rev);
+
+    reg_clr(SYS_IMP_APL_HID3, BIT(2));
+    reg_mask(SYS_IMP_APL_HID3, GENMASK(62, 56), BIT(59));
+    reg_set(SYS_IMP_APL_HID9, BIT(17));
+
+    reg_mask(SYS_IMP_APL_HID13,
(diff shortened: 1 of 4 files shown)


src/soc.h
```diff
@@ -39,8 +39,12 @@
 #define EARLY_UART_BASE 0x39b200000
 #elif TARGET == T8112
 #define EARLY_UART_BASE 0x235200000
+#elif TARGET == T8122
+#define EARLY_UART_BASE 0x2a1200000
 #elif TARGET == T8132
 #define EARLY_UART_BASE 0x3ad200000
+#elif TARGET == T8122
+#define EARLY_UART_BASE 0x2a1200000
```
asdfugil: Duplicate

src/chickens_everest.c
```diff
@@ -36,6 +36,43 @@ static void init_common_everest(void)
     reg_clr(s3_0_c15_c2_4, BIT(0) | BIT(1) | BIT(16) | BIT(17) | BIT(18) | BIT(22));
 }
 
+void init_t8122_everest(int rev)
```
asdfugil, line 39: There is a `init_common_everest()` function, it should be used.

`init_t6030_everest()` already have the duplication problem so maybe modify that one to use init_common_everest() as well.

P.S. t6031's init sequence specifically touches 1 GHz/24 MHz timer related registers s3_1_c15_c1_5 (AGTCNTRDIR_EL1) and s3_4_c15_c14_6 (AGTCNTRDIR_EL12) when the whole M3-series has them but given the state of things for this PR it's fine to leave that as is and change them later.

svenpeter42 (approved):

asdfugil (approved):

asdfugil:
Looks good, with the AGTCNTRDIR registers set to zero XNU won't boot correctly but that's the job of the HV code and a story for another time.

PR #522, opened 2025-12-18
smc: Boot SMC on A11 and T2

asdfugil:
- Turn SMC on in PMGR before trying to use it
- Turn SMC on before booting Linux so kernel can use it (SMC has no reason to be off during runtime even in sleep mode because it handles the battery, power supply and other stuff that needs to stay on)

### DIFF
--- a/src/kboot.c
+++ b/src/kboot.c
@@ -21,6 +21,7 @@
 #include "pmgr.h"
 #include "sep.h"
 #include "sio.h"
+#include "smc.h"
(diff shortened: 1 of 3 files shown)


asdfugil:
I do not think this is the right approach anymore. I should start it in Linux especially since I already need a different compatible for a different list of MFD devices.

asdfugil:
Starting SMC in Linux also potentially allows newer firmware to be used.

PR #523, opened 2025-12-18
pre-M1 mailbox support

asdfugil:
- Support T8015's ASC mailboxes as well as "AKF" mailboxes
- Make the SEP interface work with AKF mailboxes


asdfugil:
I have started work on rtkit over these 64-bit mailboxes which needed changes to the mailbox code.
I will PR again later.

PR #525, opened 2025-12-18
fix rustc 1.92.0 warnings

jfbortolotti:
Removing
warning: the feature `new_zeroed_alloc` has been stable since 1.92.0 and no longer requires an attribute to enable
in rust/src/lib.rs


WhatAmISupposedToPutHere:
Wouldn't that cause our MSRV to be 1.92?

PR #533, opened 2026-01-18
m1n1: firmware: fill in missing firmware versions

shizmob:
After verification of consistency with current data and own observations, data taken from TheAppleWiki[1].

As there's also RC versions in the table now, bump down the beta version placeholder from `99` to `98` and use `99` for RC versions. When the same iBoot version is used on multiple versions, the lowest non-beta/RC version is chosen for display.

Some version checks in e.g. `isp.c` and `kboot.c` might need expanding to cover the newly added versions.

[1]: https://theapplewiki.com/wiki/IBoot_(Bootloader)#macOS


jannau (approved):

jannau:
As for the checks in specific files those are just supposed to cover firmware versions supported by the installer and/or kernel. This are intentionally just a few selected ones (ideally just a single one per major macos release and we might want/have to skip macos 15).
No issue of printing all firmware correctly though.

PR #536, opened 2026-01-21
T8132 smp

yuyuyureka:
According to the discussion [here](https://oftc.catirclogs.org/asahi-re/2025-04-06#34157602), the RVBAR at cpu_impl_reg+0x00 is already set to the m1n1 entrypoint and locked by iBoot _for all cores_ on M4.

Indeed, I can see the following values:

m1n1 base: `0x100034f4000`
E-Cores cpu_impl_reg at `0x210X50000`: `0x100100034f4001`
P-Cores cpu_impl_reg at `0x211X50000`: `0x1100100034f4001`

And voila, when skipping the writes when the value is already correct (which previously produced SErrors when writing to the P-Core cpu_impl_reg), `smp_start_secondaries()` now works as expected.

### DIFF
--- a/proxyclient/m1n1/hv/__init__.py
+++ b/proxyclient/m1n1/hv/__init__.py
@@ -1785,8 +1785,13 @@ class HV(Reloadable):
             if cpu.state == "running":
                 continue
             addr, size = cpu.cpu_impl_reg
-            print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
-            self.p.write64(addr, rvbar)
+            val = self.p.read64(addr)
+            if val & 0xffff_ffff_f000 != rvbar:
+                if val & 1:
+                    raise Exception(f"RVBAR for {cpu.name} (={val:x}) is locked but differs from entry point (={rvbar:x})")
+
+                print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
+                self.p.write64(addr, rvbar)
 
     def _load_macho_symbols(self):
         self.symbol_dict = self.macho.symbols
--- a/proxyclient/tools/chainload.py
+++ b/proxyclient/tools/chainload.py
@@ -99,8 +99,13 @@ for cpu in u.adt["cpus"]:
     if cpu.state == "running":
         continue
     addr, size = cpu.cpu_impl_reg
-    print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
-    p.write64(addr, rvbar)
+    val = p.read64(addr)
+    if val & 0xffff_ffff_f000 != rvbar:
+        if val & 1:
+            raise Exception(f"RVBAR for {cpu.name} (={val:x}) is locked but differs from entry point (={rvbar:x})")
+
+        print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
+        p.write64(addr, rvbar)
 
 u.push_adt()
 
--- a/src/smp.c
+++ b/src/smp.c
@@ -267,6 +267,7 @@ void smp_start_secondaries(void)
             break;
         case T8112:
         case T8122:
+        case T8132:
             cpu_start_off = CPU_START_OFF_T8112;
             break;
         case T6020:
--- a/src/smp.c
+++ b/src/smp.c
@@ -24,6 +24,9 @@
 #define CPU_REG_CLUSTER GENMASK(10, 8)
 #define CPU_REG_DIE     GENMASK(14, 11)
 
+#define RVBAR_LOCK BIT(0)
+#define RVBAR_ADDR GENMASK(47, 12)
+
 struct spin_table {
     u64 mpidr;
     u64 flag;
@@ -123,6 +126,14 @@ static void smp_start_cpu(int index, int die, int cluster, int core, u64 impl, u
 
     printf("Starting CPU %d (%d:%d:%d)... ", index, die, cluster, core);
 
+    u64 rvbar_value = read64(impl);
+    bool write_rvbar = FIELD_GET(RVBAR_ADDR, rvbar_value) != (u64)_vectors_start;
+    if (FIELD_GET(RVBAR_LOCK, rvbar_value) && write_rvbar) {
+        printf("Failed! \n    RVBAR (=0x%lx) is locked and differs from entry point (=0x%lx)\n",
+               rvbar_value, (u64)_vectors_start);
+        return;
+    }
+
     memset(&spin_table[index], 0, sizeof(struct spin_table));
 
     target_cpu = index;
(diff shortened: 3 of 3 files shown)


proxyclient/tools/chainload.py
```diff
@@ -100,6 +100,15 @@ def remove_oslog(node):
...
     print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
+    # On some platforms (M4) iBoot already sets this
+    # to the m1n1 entrypoint and locks the register.
+    # Skip the write if the value is already correct.
+    val = p.read64(addr)
+    if val & 0xfffffffff000 == rvbar:
+        continue
+    if val & 1:
+        print(f"{hex(val & 0xfffffffff000)} != {hex(rvbar)}")
+        print("The cpu_impl_reg is already locked, this might fail...")
```
svenpeter42: I'd just throw an exception here or even check for this earlier and refuse to even start.

src/smp.c
```diff
@@ -140,7 +140,9 @@ static void smp_start_cpu(int index, int die, int cluster, int core, u64 impl, u
 
     sysop("dsb sy");
 
-    write64(impl, (u64)_vectors_start);
+    if ((read64(impl) & 0xfffffffff000) != (u64)_vectors_start) {
+        write64(impl, (u64)_vectors_start);
```
svenpeter42: Same here, if the register is locked and doesn't contain the right address just fail earlier

svenpeter42 (changes requested):

proxyclient/tools/chainload.py
```diff
@@ -100,7 +100,16 @@ def remove_oslog(node):
         continue
     addr, size = cpu.cpu_impl_reg
     print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
-    p.write64(addr, rvbar)
+
+    val = p.read64(addr)
+    locked = val & 1
+    do_write = val & 0xfffffffff000 == rvbar:
```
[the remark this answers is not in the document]
yuyuyureka: Thx

yuyuyureka:
I tested the following on my M4 Mac Mini:

- Installed m1n1 using kmutil
- The uartproxy appears
- I can chainload m1n1 using proxyclient/tools/chainload.py

IMO this PR is ready to be merged, as it solves the problem of causing the SError in smp init (for example when chainloading).

xarkes:
Tested on my M4 Macbook Air, chainloading works properly and I don't hit the Python Exception anymore.

abrender:
> @abrender would you mind testing it again?

Sorry for the delayed response - I no longer have access to the M4 to test :(

src/smp.c
```diff
@@ -123,6 +123,15 @@ static void smp_start_cpu(int index, int die, int cluster, int core, u64 impl, u
 
     printf("Starting CPU %d (%d:%d:%d)... ", index, die, cluster, core);
 
+    u64 rvbar_value = read64(impl);
+    bool rvbar_locked = rvbar_value & 1;
+    bool write_rvbar = (rvbar_value & 0xfffffffff000) != (u64)_vectors_start;
```
jannau: please use `~0xfffUL`.
`update_rvbar` sounds to me clearer than `write_rvbar`

proxyclient/m1n1/hv/__init__.py
```diff
@@ -1786,7 +1786,16 @@ def remove_oslog(node):
                 continue
             addr, size = cpu.cpu_impl_reg
             print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
-            self.p.write64(addr, rvbar)
+
+            val = p.read64(addr)
+            locked = val & 1
+            do_write = val & 0xfffffffff000 != rvbar
```
jannau: I think
```
if val & ~0xfff != rvbar:
    if val & 1:
        raise Exception(f"RVBAR for {cpu.name} is locked but differs from entry point")

    p.write64(addr, rvbar)
```
is easier to follow than the ifs with two booleans. `print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")` should stay directly above the write.

Not sure if we shouldn't just drop this. I think this is a leftover from old times before we updated in-place

proxyclient/tools/chainload.py
```diff
@@ -100,7 +100,16 @@ def remove_oslog(node):
         continue
     addr, size = cpu.cpu_impl_reg
     print(f"  {cpu.name}: [0x{addr:x}] = 0x{rvbar:x}")
-    p.write64(addr, rvbar)
+
+    val = p.read64(addr)
+    locked = val & 1
+    do_write = val & 0xfffffffff000 != rvbar
```
jannau: same here

src/smp.c
```diff
@@ -123,6 +123,15 @@ static void smp_start_cpu(int index, int die, int cluster, int core, u64 impl, u
...
+    if (rvbar_locked && write_rvbar) {
+        printf("RVBAR is locked and does not already contain start address:\n  0x%lx != 0x%lx\n",
```
jannau: Please prefix it with `Failed! \n    ` to make it clear that starting the secondary CPU failed.
"and differs from entry point" sound imo nicer. Printing the values could make clear which one is which

jannau (commented):
functionally ok now, sorry for the nitpicking

src/smp.c
yuyuyureka: `~0xfffUL` would not work. According to the notes in the PR description, the value I read from the register was `0x100100034f4001` for E-cores and `0x1100100034f4001` for P-Cores, while the m1n1 base is `0x100034f4000`.

So, we want to skip if the last 48 bits (except the last 12 bits) are already set to the expected value, since the first 16 bits may contain some other information and are not used for the actual jump address.

jannau: ok, please use defines then
```
#define RVBAR_LOCK BIT(0)
#define RVBAR_ADDR GENMASK(47, 12)
```

I don't think we need `rvbar_locked` then since `FIELD_GET(RVBAR_LOCK, val)` is as descriptive.

proxyclient/m1n1/hv/__init__.py
yuyuyureka: Since ~0xfff does not work, is it fine if I use 0x00000fffffffff000 (a full 64-bit value) as mask to be more clear? I didn't find an equivalent to GENMASK on python.

jannau: The equivalent would be `Register64` but I was fine with unchanged python code. Maybe write it as `0xffff_ffff_f000` as that makes it more readable

src/smp.c
yuyuyureka: Done

src/smp.c
yuyuyureka: Done

proxyclient/m1n1/hv/__init__.py
yuyuyureka: > Maybe write it as 0xffff_ffff_f000 as that makes it more readable

done

yuyuyureka:
Tested chainload on T8140

jannau (approved):

PR #539, opened 2026-01-30
build: support building using MacPorts LLVM on Darwin

shizmob:
This PR implements support for building on Darwin using MacPorts, by first querying if `llvm-config` is available and using that. This should be a generic approach that works on multiple kinds of environments.
The Homebrew approach is also changed to use Homebrew's `llvm-config`.


svenpeter42:
feel free to add CI, it should be free since the repo is public and we just use the default runners

PR #562, opened 2026-04-10
shell: handle PermissionError when reading history file

a-ramses:
## Summary
- On macOS, extended attributes (e.g. `com.apple.provenance`) can cause `PermissionError` when reading `~/.m1n1-history`, even when the user has appropriate file permissions.
- Catch `PermissionError` alongside `FileNotFoundError` so the REPL doesn't crash on startup.

### DIFF
--- a/proxyclient/m1n1/shell.py
+++ b/proxyclient/m1n1/shell.py
@@ -27,6 +27,11 @@ class HistoryConsole(code.InteractiveConsole):
                 readline.read_history_file(histfile)
             except FileNotFoundError:
                 pass
+            except PermissionError as e:
+                print(f"Failed reading history from {histfile}: {e}", file=sys.stderr)
+                if sys.platform == "darwin":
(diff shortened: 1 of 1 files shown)


jannau:
shouldn't this remain an error so that the user can fix the issue?

Commit misses Developer certificate of origin / `Signed-off-by:`

svenpeter42:
Ignoring File not Found is fine because it'll be created anyway, but we shouldn't silently ignore PermissionError. Maybe just print a helpful message how to fix those extended attributes or whatever other most common causes exist

jannau:
I think warning and carrying on is ok for the history file. Absence of the history is annoying but shouldn't be fatal.  
The commit message is confusing. `... instead of silently ignoring the error.` does not make sense. Without your change it was fatal error. Silently ignoring it was the initial change but that doesn't need to be mentioned in the commit message.

svenpeter42 (approved):

PR #563, opened 2026-04-16
smp: Allow switching the boot (primary) CPU core

jannau:
```
M3 and later Apple silicon SoCs start m1n1 on the first perfomance core.
The cluster layout remained unchanged with an efficiency core cluster
first. This causes an annoying but mostly cosmetic issue in Linux. Linux
assigns CPU index 0 to the boot CPU. This results in a mismatch of
logical and physical CPU core order. As seen with the Radxa Orion O6
this will at least generate support questions but is also annoying for
CPU pinning. It's likely to result in pinning to mixture of performance
and efficiency cores.
While functionality it should not matter to which secondary CPU core
control is passed the only expected use case is switching to CPU core 0.
```

### DIFF
--- a/src/proxy.c
+++ b/src/proxy.c
@@ -374,6 +374,9 @@ int proxy_process(ProxyRequest *request, ProxyReply *reply)
                       request->args[3], request->args[4]);
             reply->retval = smp_wait(request->args[0]);
             break;
+        case P_SMP_SWITCH_BOOT_CPU:
+            reply->retval = smp_switch_boot_cpu(request->args[0]);
+            break;
 
         case P_HEAPBLOCK_ALLOC:
             reply->retval = (u64)heapblock_alloc(request->args[0]);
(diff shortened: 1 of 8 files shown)


src/proxy.c
```diff
@@ -374,6 +374,9 @@ int proxy_process(ProxyRequest *request, ProxyReply *reply)
                       request->args[3], request->args[4]);
             reply->retval = smp_wait(request->args[0]);
             break;
+        case P_SMP_SWITCH_BOOT_CPU:
+            reply->retval = smp_wait(request->args[0]);
```
yuyuyureka: This should be smp_switch_boot_cpu?

jannau: good catch, not sure how it ended up as such. I remember editing it after pasting. I'll fix

jannau:
I've seen now some unhandled exceptions after switching the boot CPU during DT preparation so it's possible that there is something missing during initialisation of the new boot CPU

yuyuyureka:
Could we already merge the bits in src/smp.c without using it in tools/linux.py until we have figured the unhandled exceptions?

jannau:
first it needs to be rebased. I don't see this as urgent as it fixes a mostly cosmetic issue.

yuyuyureka:
I have rebased it here: https://github.com/yuyuyureka/m1n1/commits/switch_boot_cpu/

PR #566, opened 2026-05-04
convert -> magick

davide125:
ImageMagick 7 and later deprecate the convert command, which now prints a warning. Use the new magick one instead, which takes the same arguments.


jannau (approved):
I've seen that as well and was planning to make a similar change. I was planning to use `magick convert` though since I think that is clearer. Change is ok in either case.

davide125:
`magick convert` also triggers the deprecation warning on my system for some reason, so I'll stick with `magick` for now.

PR #568, opened 2026-05-05
rust: bump uuid to 1.23 and switch to using a workspace

davide125:
Bump to what's shipped in Fedora


flokli:
I can confirm this fixes the build for me when using nixpkgs tooling, which currently fails like this on `main`:

```
error: failed to get `uuid` as a dependency of package `rust v0.1.0 (/build/source/rust)`

Caused by:
  failed to load source for dependency `uuid`

Caused by:
  Unable to update registry `crates-io`

Caused by:
  failed to update replaced source registry `crates-io`

Caused by:
  failed to load checksum `.cargo-checksum.json` of rust v0.1.0

Caused by:
  failed to read `/build/source/rust/.cargo-checksum.json`

Caused by:
  No such file or directory (os error 2)
make: *** [Makefile:216: build/librust.a] Error 101
make: *** Waiting for unfinished jobs....
```

nixpkgs commit: https://github.com/flokli/nixpkgs/commit/89b9af1bc6680764b6ec125afb95fa85ffdbb91b

rust/Cargo.toml
```diff
@@ -11,8 +11,16 @@ crate-type = [ "staticlib" ]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [dependencies]
-fatfs = { path = "vendor/rust-fatfs", default-features = false, features = ["lfn", "alloc"], optional = true }
-uuid = { version = "1.7.0", default-features = false, optional = true }
+fatfs = { workspace = true, optional = true }
```
jannau, line 14: keep `default-features = false, features = ["lfn", "alloc"]` I assume drooping those is responsible for large number of added dependencies in `rust/Cargo.lock`. This should result in a broken 1st stage / chainload enabled m1n1

jannau (changes requested):

rust/Cargo.toml
davide125: It's actually there down below, and adding it back here doesn't seem to make a difference. Agree that in the current form this isn't something we can ship. I'm going to let this sit for now and factor out the uuid bump in a separate PR so we can get that one merged at least.

PR #571, opened 2026-05-06
rust: adt: Handle missing 'ranges' prop

sofus13:
Break early instead of returning an error.
Mirrors original C behavior.

Without this, m1n1 fails to create the nvram partition

### DIFF
--- a/rust/src/adt.rs
+++ b/rust/src/adt.rs
@@ -211,7 +211,11 @@ pub fn get_reg_container(
             }
         };
 
-        let ranges = node.named_prop("ranges")?;
+        let ranges = match node.named_prop("ranges") {
+            Ok(r) => r,
+            Err(AdtError::NotFound) => break,
+            Err(e) => return Err(e),
+        };
 
         let paddr_cells = parent.named_prop("#address-cells")?.u32()?;
 


WhatAmISupposedToPutHere:
Which device does this issue happen on?

sofus13:
j416s

WhatAmISupposedToPutHere (approved):

jannau:
the template j416sap ADT has a ranges property in the `spinor` node (parent of `anvram`) so I'd expect this to work. The change looks however correct as in that it restores behavior of `adt.c` which had the break if no "ranges" property was found. I wonder if it would be worth to just break on `AdtError::NotFound` and handle other errors still as errors.

I'm also not sure why we loop here at all. Using the parent's parent "ranges" does not make sense. It only happens if the current "ranges" has no matching range. I don't think this inconsistency makes sense. Unfortunately I don't trust the soundness of the ADT enough to exclude this is actually used somewhere.

sofus13:
I think the `spinor` "ranges" property has `child_addr=0` and therefore does not break.

I agree with the error handling and updated it to do that

jannau (approved):
thanks

PR #572, opened 2026-05-06
Initial support for T8140

yuyuyureka:
Adds EARLY_UART_BASE, smp start offset and preliminary features.


src/chickens.c
```diff
@@ -66,6 +66,13 @@ const struct midr_part_features features_a11 = {
     .fast_ipi = true,
 };
 
+// XXX figure out what features are actually available on A18
+const struct midr_part_features features_a18 = {
```
svenpeter42: As long as the settings are the same as m4 I wouldn't introduce another struct here but just reuse the one for m4

svenpeter42 (changes requested):

src/chickens.c
yuyuyureka: Done

jannau (approved):

svenpeter42 (approved):

PR #575, opened 2026-05-06
smp: Compare non-shifted RVBAR against _vectors_start

jannau:
RVBAR re-purposes the 12 least signigficant bits for other purposes. Do not uses FIELD_GET() since that will discard the least significant bits via a right shift.

Fixes: ed111502 ("smp: only write rvbar if needed")


yuyuyureka (approved):

jannau:
I'll revert the changes from #536 instead. There was wrong understanding how BIT(0) locks the `cpu-impl-reg`. On M1, M2 and M3 it prevents writes when the CPU is running. If the CPU is offline it can be modified at will.

PR #577, opened 2026-05-06
smp: only write rvbar when apple_sysregs_unlocked

yuyuyureka:


svenpeter42:
LGTM, but let's rename `cyc_ovrd` to something like `apple_sysregs_unlocked`

svenpeter42 (approved):

src/utils.c
```diff
@@ -225,8 +225,8 @@ u64 top_of_memory_alloc(size_t size)
 extern void _deep_wfi_helper(void);
 void deep_wfi(void)
 {
-    u64 cyc_ovrd;
-    const bool cpu_feat_cyc_ovrd = cpu_features->cyc_ovrd;
+    u64 apple_sysregs_unlocked;
```
svenpeter42: search and replace went wrong here for the u64, this should stay as cyc_ovrd because it's not the feature flag but the actual stored value of that sysreg

svenpeter42 (changes requested):

src/utils.c
yuyuyureka: Thanks for catching that. Is there is a reason why the value is copied to this local variable and not referred to as cpu_features->apple_sysregs_unlocked directly?

svenpeter42: https://github.com/AsahiLinux/m1n1/commit/0478c7989001f4e084a672635579ccd60f2c9d48 looks like that was added to silence a compiler warning

PR #579, opened 2026-05-10
wdt: check ADT compatible strings for secondary watchdog

flokli:
Comparing the offset between the first and third watchdog address in the ADT is not sufficient, it's not 0x8008 for the M4 Pro and M5 Max at least.

Instead, apply the secondary watchdog logic for all boards where wdt_version is 2 or 3.

Also update the printf statements, so it becomes more clear if a secondary watchdog gets disabled or not.

### DIFF
--- a/src/wdt.c
+++ b/src/wdt.c
@@ -9,8 +9,6 @@
 #define WDT_ALARM 0x14
 #define WDT_CTL   0x1c
 
-#define WDT_2ND_OFFSET 0x8008
-
 static u64 wdt_base = 0;
 
 void wdt_disable(void)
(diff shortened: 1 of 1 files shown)


src/wdt.c
```diff
@@ -28,19 +26,36 @@ void wdt_disable(void)
...
+    printf("Primary WDT register @ 0x%lx\n", wdt_base);
+    write32(wdt_base + WDT_CTL, 0);
+    printf("Primary WDT disabled\n");
+
+    // disable secondary watchdog enabled on M4 / M5 / A18 Pro with macOS 26
+    int path2[8];
+    int node_arm_io = adt_path_offset_trace(adt, "/arm-io", path2);
+
+    if (node < 0) {
+        printf("AIC node not found!\n");
```
asdfugil: This is not the AIC, and this doesn't check `node_arm_io` either.

src/wdt.c
```diff
@@ -28,19 +26,36 @@ void wdt_disable(void)
...
     }
 
-    printf("WDT registers @ 0x%lx (0x%lx)\n", wdt_base, wdt_2nd);
+    if (
+      adt_is_compatible(adt, node_arm_io, "arm-io,t8132") ||
+      adt_is_compatible(adt, node_arm_io, "arm-io,t6040") ||
+      adt_is_compatible(adt, node_arm_io, "arm-io,t6041") ||
+      adt_is_compatible(adt, node_arm_io, "arm-io,t8140") ||
+      adt_is_compatible(adt, node_arm_io, "arm-io,t8142") ||
+      adt_is_compatible(adt, node_arm_io, "arm-io,t6050")) {
```
asdfugil: The existing secondary watchdog logic triggers when the secondary watchdog at 0x8008 exists, that is to say, the baseline M-series SoCs starting from T8103. (not Pro/Max/Ultra)

It is more future-proof to check the `/arm-io/wdt/wdt-version` property. It is 2 (since T8103) if the secondary watchdog and reg[2] exists, 1 otherwise (previous SoCs).

This way, the secondary watchdog is disarmed when it exists, and SoCs without them do not get bogus error messages.

asdfugil (changes requested):

src/wdt.c
flokli: As suggested, updated the code to look at `wdt-version` instead. PTAL.

src/wdt.c
```diff
@@ -28,19 +26,26 @@ void wdt_disable(void)
...
+    write32(wdt_base + WDT_CTL, 0);
+    printf("Primary WDT disabled\n");
+
+    // disable secondary watchdog if wdt-version is 2
+    u32 wdt_version;
+    if (ADT_GETPROP(adt, node, "wdt-version", &wdt_version) < 0)
+        return;
 
-    printf("WDT registers @ 0x%lx (0x%lx)\n", wdt_base, wdt_2nd);
+    if (wdt_version == 2) {
```
yuyuyureka: J714c (MacBook Pro 14" M5 Max) has the secondary watchdog, and the disable code works for it, but it has wdt-version = 3

flokli: I updated the check to match on wdt-version 2 and 3.

asdfugil (approved):
Confirmed working without bogus messages on SoCs with wdt-version == 1 (T8012 in this case)

svenpeter42 (approved):

jannau (approved):

PR #583, opened 2026-05-29
hv: macOS EL1 fixups for M3 devices

jannau:
Replace known EL2-only mrs/msr instructions out of MachO payloads and manually initialize them to the expected values on guest startup.


jannau:
dropped the `HACR_EL2` intialising/copying as that shouldn't matter. It's still patched out of the MachO text section

svenpeter42 (approved):

PR #587, opened 2026-06-02
experiments/cpu_pstate_latencies: Fix with 1GHz clock

yuyuyureka:
signal_and_write waits 2048 24MHz cycles ~= 85μs for the cpu under test
to start logging.
    
This commit fixes 2 issues regarding the wait:
    
1. The number of cycles was not accurate on systems which use a 1GHz
   clock for CNTPCT_EL0. Changed it to dynamically calculate the cycles
   for a 100μs wait (to be sure) based on tfreq.
2. The loop expected the timer value to land exactly on the target,
   which usually worked with a clock which is much slower than the CPU
   freq. But with both the CPU and CNTPCT_EL0 being clocked close to
   1GHz there is some chance that we go past the value. Changed it to
   stop if we are past the target value.

### DIFF
--- a/proxyclient/experiments/cpu_pstate_latencies.py
+++ b/proxyclient/experiments/cpu_pstate_latencies.py
@@ -47,6 +47,9 @@ elif chip_id in (0x8112, 0x6020, 0x6021, 0x6022):
 
 code = u.malloc(0x1000)
 
+make_imm = lambda x : x & (0b11111 << (x.bit_length() - 5))
+eigthy_six_us = make_imm(round(87 * (tfreq / 1_000_000)))
+
 util = asm.ARMAsm(f"""
 bench:
     mrs x1, CNTPCT_EL0
@@ -61,11 +64,11 @@ bench:
 signal_and_write:
     sev
     mrs x2, CNTPCT_EL0
-    add x2, x2, #0x800
+    add x2, x2, #{hex(eigthy_six_us)}
 1:
     mrs x3, CNTPCT_EL0
-    sub x4, x3, x2
-    cbnz x4, 1b
+    cmp x3, x2
+    blt 1b
     str x1, [x0]
     mov x0, x3
     ret


proxyclient/experiments/cpu_pstate_latencies.py
```diff
@@ -38,6 +38,8 @@
 
 code = u.malloc(0x1000)
 
+hundred_microsecs = round(100/(1/(tfreq/1_000_000)))
```
jannau: please use `round(100 * (tfreq / 1_000_000))`, 96 or 128 are nicer values but neither is good enough (results in a multiple of 4096 for 1 GHz counter freq) to still use the immediate add

proxyclient/experiments/cpu_pstate_latencies.py
```diff
@@ -51,12 +53,14 @@
 
 signal_and_write:
     sev
+    ldr x5, ={hex(hundred_microsecs)}
```
jannau: `mov x5, hex(hundred_microsecs)` hex shouldn't be necessary

proxyclient/experiments/cpu_pstate_latencies.py
```diff
@@ -51,12 +53,14 @@
...
     mrs x2, CNTPCT_EL0
-    add x2, x2, #0x800
+    add x2, x2, x5
 1:
     mrs x3, CNTPCT_EL0
     sub x4, x3, x2
-    cbnz x4, 1b
+    cmp x4, #0
+    blt 1b
```
jannau, line 71: this can become with the preceeding `sub`:
```
cmp x3, x4
blt 1b
````

**Review by jannau — COMMENTED** [review]
Would it make more sense to run this with counter scaling disabled? 
```
        u.msr(AGTCNTRDIR_EL1, 3)
        u.msr(AGTCNTRDIR_EL12, 3)
```
at the beginning of the file?

The change is still good. M4 and A18 Pro supposedly already come out of iboot with `CNTFREQ_EL0 == 1_000_000_000`.

I question my decision to apply the scaling in chickens. It might be better to move that to kboot. My concern was a varying CNTFREQ during m1n1's execution. But we already have that for running macOS in the hypervisor.

**Comment by yuyuyureka** [comment]
> Would it make more sense to run this with counter scaling disabled?

On reason against doing this is that it is not allowed on M4:

```
>>> mrs(AGTCNTRDIR_EL1)                                                           
0x0                                                                                
>>> msr(AGTCNTRDIR_EL1, 3)                                                         
TTY> Exception: SYNC                                                               
TTY> Exception taken from EL2h                                                     
TTY> Running in EL2                                                                
TTY> MPIDR: 0x80010100                                                             
TTY> Registers: (@0x10002a47c50)                                                   
TTY>   x0-x3: 0000000000000003 0000000000000000 0000000000000000 0000000000000000  
TTY>   x4-x7: 0000000000000000 00000d000c804000 0000010002a47f18 0000010002a0f000  
TTY>  x8-x11: 0000010002a47f40 0000000000000001 0000000000000000 0000000000000000  
TTY> x12-x15: 0000000000000000 0000000000000000 0000000000000000 0000000000000000  
TTY> x16-x19: 0000000000000000 0000000000000000 0000000000000000 0000000000000001  
TTY> x20-x23: 0000010002994000 0000010002a1ad08 0000010002a1ad08 0000000000aa55ff  
TTY> x24-x27: 0000000000000000 0000010002a47f10 0000000000000001 0000010002a0f000  
TTY> x28-x30: 00000000addedbad 0000010002a47e50 0000010002956a94                   
TTY> PC:       0xd000c804000 (rel: 0xc0009ed8000)                                  
TTY> SP:       0x10002a47d50                                                       
TTY> SPSR:     0x80000809                                                          
TTY> FAR:      0x0                                                                 
TTY> ESR:      0x2000000 (unknown)                                                 
TTY> L2C_ERR_STS: 0x0                                                              
TTY> L2C_ERR_ADR: 0x0                                                              
TTY> L2C_ERR_INF: 0x0                                                              
TTY> Exception: SYNC                                                               
TTY> Exception taken from EL2h  
```

File: proxyclient/experiments/cpu_pstate_latencies.py (hunk 2 above)
Comment (jannau) [reply proxyclient/experiments/cpu_pstate_latencies.py]: err. I can't do math. the value exceeds 16 bits for 1 GHz counter freq. We could still use `one_microsec` or `three_micosecs` with mov and then use `add x2, x2, x5, lsl #7` (128 ms) or `add x2, x2, x5, lsl #5` (96 ms)

Comment (yuyuyureka) [reply proxyclient/experiments/cpu_pstate_latencies.py]: 96μs in 1GHz is 0x17700 cycles which is too big for mov, no?

File: proxyclient/experiments/cpu_pstate_latencies.py (hunk 1 above)
Comment (jannau) [reply proxyclient/experiments/cpu_pstate_latencies.py]:
```
make_imm = lambda x : x & (0b11111 << (x.bit_length() - 5))
eigthy_six_us = make_imm(round(87 * (tfreq / 1_000_000)))

...


    add x2, x2, #{hex(eigthy_six_us)}
```
ends up with the current period of 85 - 86 microseconds for both cases and works for 24 MHz and 1 GHz

proxyclient/experiments/cpu_pstate_latencies.py
yuyuyureka: `cmp x3, x2`. x4 is not used for anything without the sub.

proxyclient/experiments/cpu_pstate_latencies.py
yuyuyureka: Thanks, TIL the immediate value can have a shift

yuyuyureka:
Updated with your suggested changes

jannau (approved):

PR #594, opened 2026-06-06
tps6598x: add spmi transport

yuyuyureka:


src/tps6598x.c
```diff
@@ -50,29 +70,78 @@ void tps6598x_shutdown(tps6598x_dev_t *dev)
...
+    u8 val = ~reg;
+    if (spmi_reg0_write(dev->spmi, dev->addr, reg) < 0)
+        return -1;
+
+    while (val != reg) {
+        if (spmi_ext_read(dev->spmi, dev->addr, 0, &val, 1) < 0)
+            return -1;
+        if (val == reg)
+            break;
+        if (val != (reg | 0x80))
```
svenpeter42: do you know what 0x80 means here? please #define these magic numbers here (and further below) if possilbe

src/tps6598x.c
```diff
@@ -50,29 +70,78 @@ void tps6598x_shutdown(tps6598x_dev_t *dev)
...
+
+static int tps6598x_write_reg(tps6598x_dev_t *dev, const u8 reg, const u8 *data, size_t len)
+{
+    if (dev->i2c)
+        return i2c_smbus_write(dev->i2c, dev->addr, reg, data, len);
+
+    if (dev->spmi) {
+        if (tps6598x_spmi_select_reg(dev, reg) < 0)
+            return -1;
+        if (spmi_ext_write(dev->spmi, dev->addr, 0xa0, data, len) < 0)
```
svenpeter42: what's 0xa0?

src/tps6598x.c
```diff
@@ -50,29 +70,78 @@ void tps6598x_shutdown(tps6598x_dev_t *dev)
...
+
+static int tps6598x_read_reg(tps6598x_dev_t *dev, const u8 reg, u8 *data, size_t len)
+{
+    if (dev->i2c)
+        return i2c_smbus_read(dev->i2c, dev->addr, reg, data, len);
+
+    if (dev->spmi) {
+        if (tps6598x_spmi_select_reg(dev, reg) < 0)
+            return -1;
+        if (spmi_ext_read(dev->spmi, dev->addr, 0x20, data, len) < 0)
```
svenpeter42: what's 0x20?

svenpeter42 (changes requested):
just a quick glance, will take a look at the rest probably on the weekend

yuyuyureka:
Updated with defines for the offsets. According to https://asahilinux.org/docs/hw/peripherals/ace3/, 0xa0 is an alias to 0x20, so I'm re-using that as TPS_SPMI_REG_DATA.

yuyuyureka:
Marked as draft because something is weird with commands which take >1 byte params, like LOCK

yuyuyureka:
Rebased and marked ready, as USE_DEBUG_USB is working on the J700 / MacBook Neo!

svenpeter42:
i'm a bit confused why there are rust changes in this PR now, I don't see them used anywhere.

WhatAmISupposedToPutHere:
> i'm a bit confused why there are rust changes in this PR now, I don't see them used anywhere.

Those look like they were accidentally pulled from https://github.com/AsahiLinux/m1n1/pull/614 somehow

src/tps6598x.c
```diff
@@ -265,55 +344,91 @@ int tps6598x_enable_debugusb(void)
...
-            ret = snprintf(hpm_path, sizeof(hpm_path), "/arm-io/%s/%s/%s", adt_get_name(adt, node),
-                           adt_get_name(adt, mngr_node), name);
-            if (ret < 0 || (size_t)ret >= sizeof(hpm_path))
-                continue;
-
-            found = true;
+            int it = mngr_node;
+            ADT_FOREACH_CHILD(adt, it)
+            {
+                if (adt_is_compatible(adt, it, "usbc,cd3217"))
```
svenpeter42: maybe it's too late or too hot but I can't really follow what's going on here anymore. can you maybe add a few comments? or maybe split this out into multiple functions? 

I think this should be !adt_is_compatible though

yuyuyureka:
The Rust changes were required for adt_is_compatible not to crash when running on nodes without a compatible prop (which is the case on some first-children of i2c nodes on M4+ devices). They're now already merged into main (through #614).

I will make this into a draft again, because I need to add a check for spmi gen4+ first, otherwise it will crash M5+ devices while trying to initialize the hpm.

yuyuyureka:
Rebased on top of #636, making many of the changes in usb.c and enter_debugusb unnecessary. Added comments so hopefully the code is easier to understand.

PR #599, opened 2026-06-14
tps6598x: Add compile time config to enable debugusb

jannau:
Add a compile time switch to enable debugusb on the DFU USB-C port for devices with I2C based USB-C port controller.

### DIFF
--- a/src/tps6598x.c
+++ b/src/tps6598x.c
@@ -5,9 +5,11 @@
 #include "i2c.h"
 #include "iodev.h"
 #include "malloc.h"
+#include "string.h"
 #include "types.h"
 #include "utils.h"
 
+#define TPS_REG_MODE        0x03
 #define TPS_REG_CMD1        0x08
 #define TPS_REG_DATA1       0x09
 #define TPS_REG_INT_EVENT1  0x14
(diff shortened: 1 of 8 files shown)


WhatAmISupposedToPutHere (approved):

src/tps6598x.c
```diff
@@ -170,3 +193,140 @@ int tps6598x_powerup(tps6598x_dev_t *dev)
...
+    char i2c_path[64] = {0};
+    bool found = false;
+    int node;
+    int ret;
+
+    node = adt_path_offset(adt, "/arm-io");
+    if (node < 0)
+        return -1;
+
+    ADT_FOREACH_CHILD(adt, node)
```
yuyuyureka, line 266: A nit: why do we enumerate here, while we assume in usb.c that hpms are on i2c0 (and even on the Mac Studio which has hpms on i2c3 I'd still assume hpm0 is on i2c0)? Seems like this could be a bit simpler under this assumption.

jannau: no, the Mac Pro has hpm0 on i2c3.
This is mostly a left-over from testing. I had a proxy command to switch arbitrary ports to debugusb (to test the i2c communication).
For the intended purpose hardcoding `/arm-io/i2c0/hpmBusManager/hpm0` and `/arm-io/i2c3/hpmBusManager0/hpm0` is likely a better idea. Not that this is important as all i2c cd321x devices have working usb device mode

jannau: keeping this as is. We can copy the code for usb/i2c and simplify this later

PR #601, opened 2026-06-15
tools/kmutil.py: Add kmutil configure-boot script server

jannau:
Add a Python http.server based program to serve a script to run `bputil configure-boot …` with embedded m1n1.bin in recovery on an Apple silicon Mac.
This replaces the 'build/kmutil' target which has to be remade when. The main advantage of this server is that it serves the script for every path. This saves a few keystrokes on the target device as it can be executed via `sh <(curl $HOST:8000)`.


tools/kmutil.py
```diff
@@ -0,0 +1,118 @@
...
+# SPDX-License-Identifier: MIT
+
+from base64 import b64encode
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+
+SCRIPT_HEADER = b'''#!/bin/sh
+#######################################################################
+# kmutil configure-boot script with embedded m1n1.bin for easy setup
+#
+# host dependencies: uuencode (sharutils)
```
flokli: This is only a dependency on the target (as `bless`, `diskutil`, `plutil` and `kmutil` are, which are not documented).

We invokend `base64` in the makefile on the host, and now we base64-encode from python, so I'd argue we can just drop this line entirely.

tools/kmutil.py
```diff
@@ -0,0 +1,118 @@
...
+                self.wfile.write(SCRIPT_TRAILER)
+        except FileNotFoundError:
+            self.send_response(404)
+            self.end_headers()
+
+
+if __name__ == '__main__':
+    import argparse, contextlib, pathlib, socket
+
+    parser = argparse.ArgumentParser(color=True)
```
flokli:
```suggestion
    parser = argparse.ArgumentParser()
```

`color=True` is python 3.14 only, and in my dev setup (NixOS 26.05) python still defaults to 3.13.

flokli: removed, thanks!

tools/kmutil.py
flokli: removed, thanks!

tools/kmutil.py
jannau: fixed

flokli:
Successfully used this to flash a new build (main plus this cherry-picked on top) to my J773s, so LGTM!

tools/kmutil.py
jannau: `uuencode` is not used on the target but was initially on the host. The target dependencies are not listed because macOS recovery is a known target.

tools/kmutil.py
```diff
@@ -0,0 +1,116 @@
...
+            return super().server_bind()
+
+        def finish_request(self, request, client_address):
+            self.RequestHandlerClass(request, client_address, self,
+                                     payload=args.payload)
+
+    class HTTPDualStackServer(DualStackServerMixin, ThreadingHTTPServer):
+        pass
+
+    server = HTTPDualStackServer((args.bind, args.port), KmutilHTTPRequestHandler)
```
flokli, line 116: It would be helpful if this could log args.bind and args.port before starting to serve.

I didn't remember it was port 8000 when I was using this yesterday, and had to look it up.

svenpeter42 (approved):

tools/kmutil.py
jannau: done

PR #602, opened 2026-06-17
experiments/wfi: test el1 and wfit

yuyuyureka:
Removed the ability to set cyc_ovrd, since it doesn't work in el1, and can be done set like any other chicken by chainloading a modified m1n1.

### DIFF
--- a/proxyclient/experiments/wfi.py
+++ b/proxyclient/experiments/wfi.py
@@ -1,18 +1,19 @@
+#!/usr/bin/env python3
 # SPDX-License-Identifier: MIT
+
 import sys, pathlib
 sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
 
 from m1n1.setup import *
 from m1n1 import asm
 
-cyc_ovrd = False
-
 code = u.malloc(0x1000)
 c = asm.ARMAsm("""
+test_wfi:
     // enable timer interrupts
-    msr CNTP_TVAL_EL0, x0
+    msr CNTV_TVAL_EL0, x0
     mov x1, #1
-    msr CNTP_CTL_EL0, x1
+    msr CNTV_CTL_EL0, x1
 
     mov x0, #1  // canary, not saved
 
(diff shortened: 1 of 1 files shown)


proxyclient/experiments/wfi.py
jannau: please add `#!/usr/bin/env python3` and make the file executable

proxyclient/experiments/wfi.py
```diff
@@ -45,7 +35,40 @@
...
+    stp x28, x29, [sp, #-16]!
+    stp x26, x27, [sp, #-16]!
+    stp x24, x25, [sp, #-16]!
+    stp x22, x23, [sp, #-16]!
+    stp x20, x21, [sp, #-16]!
+    stp x18, x19, [sp, #-16]!
+
+    isb
+    mov x1, #-1
+    msr s0_3_c1_c0_1, x1
```
jannau: this could use a comment that it produces the encoding for `wfit`

proxyclient/experiments/wfi.py
```diff
@@ -45,7 +35,40 @@
...
-    msr CNTP_CTL_EL0, x1
+    msr CNTV_CTL_EL0, x1
+
+    ret
+
+test_wfit:
+    // enable timer interrupts
+    msr CNTV_TVAL_EL0, x0
+    mov x1, #1
+    msr CNTV_CTL_EL0, x1
```
jannau: I would have tested the timeout and not the interrupt part of `wfit`. It would be possible to do both if the `wfit` timeout is passed as parameter

yuyuyureka: The answer to this is somewhat complicated: I tested it with various timeouts. With lower timeout values (<2^40), wfit immediately continues (disregarding the timeout) and does not lose state. With higher values (>2^42) it does respect the timeout, and it would take a long time to actually wait for the timeout to happen, so I opted to pass a very large timeout and then interrupt it early to trigger the state loss.
I can still adjust the function to allow testing of different wfit timeout values, and add a test which passes a somewhat low wfit timeout value and very high timer interrupt value, but this always retains state on all of my machines.

proxyclient/experiments/wfi.py
yuyuyureka: Done

proxyclient/experiments/wfi.py
yuyuyureka: Done

proxyclient/experiments/wfi.py
yuyuyureka: Added variants "wfit persists x0 after interrupt" and "wfit persists x0 after timeout"

proxyclient/experiments/wfi.py
```diff
@@ -1,18 +1,19 @@
 # SPDX-License-Identifier: MIT
+#!/usr/bin/env python3
```
jannau, line 1: I'm almost sure shebang has to be on the first line. did you test it? `./[experiments/wfi.py` is expected to work

jannau (commented):
looks good otherwise

proxyclient/experiments/wfi.py
yuyuyureka: That's correct. Fixed.

jannau (approved):

PR #603, opened 2026-06-18
proxyclient: fetch and parse cpu_features

yuyuyureka:


proxyclient/m1n1/proxy.py
```diff
@@ -480,6 +480,34 @@ class GUARD(IntFlag):
...
+UncoreVersion = "UncoreVersion" / Enum(Int32ul,
+    UNCORE_NONE = 0,
+    UNCORE_V1 = 1,
+    UNCORE_V2 = 2,
+)
+
+CPUFeatures = Struct(
+    "sleep_mode" / SleepMode,
+    "uncore_version" / UncoreVersion,
+    "disable_dc_mva" / Int8ul,
```
jannau: this probably works in practice but `bool` in C23 has 7 padding bits with unspecified values. Not sure how to solve this elegantly.
```
'disable_dc_mva' / BitStruct('_padding' / BitsInteger(7), 'bool' / Flag)
```
would work but is a little annoying to use. We could use bit-fields to get rid of the padding and then have BitStruct for all the flags or the following:
```
"flags" / BitStruct(
 "_pad0" / BitsInteger(7),
 "disable_dc_mva" / Flag),
 "_pad1" / BitsInteger(7),
 "acc_cfg" / Flag),
....
```

src/proxy.c
```diff
@@ -54,6 +54,9 @@ int proxy_process(ProxyRequest *request, ProxyReply *reply)
         case P_GET_BOOTARGS:
             reply->retval = boot_args_addr;
             break;
+        case P_GET_CPU_FEATURES:
+            reply->retval = (u64)cpu_features;
```
jannau: while mostly harmless in this case a way to ensure struct definitions match between C m1n1 and python proxyclient would be nice. I unfortunately have no good ideas. One possibility would be to at least validate the size of structs. The easiest way to implement this would be proxyclient passing the its expected size and let `proxy.c` fail if it doesn't match

proxyclient/m1n1/proxy.py
yuyuyureka: We also have the `Bool(c)` adapter in proxyclient/m1n1/fw/common.py which afaict solves this elegantly (returning `bool(x & 1)` on access). Should we maybe move this out of the fw module and use it here too?

src/proxy.c
yuyuyureka: Relevant discussion on irc: https://oftc.catirclogs.org/asahi-dev/2026-06-15#1781537067-1781537119;

proxyclient/m1n1/proxy.py
jannau: yes, please move `bool_`. Not sure if the other adapters make sense. I'd move it to constructutils.py

src/proxy.c
jannau: I'm aware. Checking the struct size seems easy enough and prevents reading beyond the end of the struct on the C side if the m1n1 version is out of date (due to forgetting to chainload the matching version)

proxyclient/m1n1/proxy.py
yuyuyureka: Done. I hope I didn't miss any imports of Bool / bool_.

src/proxy.c
yuyuyureka: Implemented size checking and added one padding byte to the proxyutils definition to match the m1n1 side.

jannau (approved):
thanks

PR #605, opened 2026-06-19
cpufreq: M3 (T8122) support

jannau:
The cpufreq init on T8122 is probably identical to T6030 / T6031. The clearing of CLUSTER_PSTATE_M2_APSC_DIS in `t6030_features` might be just an oversight and if so should be removed.

### DIFF
--- a/src/cpufreq.c
+++ b/src/cpufreq.c
@@ -69,6 +69,7 @@ static u32 pstate_reg_to_pstate(u64 val)
         case T6030:
         case T6031:
         case T6034:
+        case T8122:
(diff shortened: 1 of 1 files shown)


src/cpufreq.c
```diff
@@ -503,6 +517,8 @@ const struct feat_t *cpufreq_get_features(void)
             return t8103_features;
         case T8112:
             return t8112_features;
+        case T8122:
```
svenpeter42, line 520: ~~duplicate from two lines above~~ Er no, I just can't read

svenpeter42 (approved):

PR #609, opened 2026-06-28
trace_pmp,pmp_init: support more devices

yuyuyureka:
- pmp_assist: include devices without notify_pmp in the list
- Add a table of base addresses including T600X, T602X, T8112, T8122, T6030, T603[14]
- pmp_init: Dynamically determine the required TGT_WR write based on a
  list of device names
- Make PMPRegTracer ASYNC because otherwise PMP init timeouts on T6034
- Limit IOVTable handling to only T600X
- Init DCAP bootarg only on T600X

### DIFF
--- a/proxyclient/experiments/pmp_init.py
+++ b/proxyclient/experiments/pmp_init.py
@@ -169,18 +169,20 @@ class PMPEp(ASCBaseEndpoint):
     @msg_handler(0x10, PMPMessage)
     def GetIOVATable(self, msg):
         self.log(f'IovaTable: {msg}')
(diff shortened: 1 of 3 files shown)


WhatAmISupposedToPutHere (approved):
i prefer .startswith, but slicing works too

yuyuyureka:
For the ones where we need to check multiple prefixes, .startswith gets very long. And I made everything into slicing, since other prefixes might be added later.

yuyuyureka:
Just found out .startswith can take a tuple of prefixes

PR #610, opened 2026-06-29
Initial support for T6050/T6051

yuyuyureka:
Adds EARLY_UART_BASE, smp start offset and preliminary features.
    
The M5 generation has new type of core. While the marketing definitions have renamed the fastest type of core to "Super Core", we stick to "P Core", matching the adt cluster-type and MPIDR.
    
The other kinds of cores in the system are however a new type of core and do not behave like E-Cores (do not have `SYS_IMP_APL_E_*`). We will call them M-Core, matching the adt cluster-type.
    
M5 Pro boots on core 0, which is an M-Core.
Currently, smp secondaries startup is not functional: The cores turn on (as can be seen in pmgr), but never start executing our code. This was tested on macOS 26.2b2 and 27.0b2.


src/chickens.c
```diff
@@ -157,6 +157,10 @@ const struct midr_part_info midr_parts[] = {
     {MIDR_PART_T6040_BRAVA_CHOP_PCORE, "M4 Pro Brava Chop (P core)", NULL, &features_m4},
     {MIDR_PART_T8140_TAHITI_ECORE, "A18 Pro Tahiti (E core)", NULL, &features_m4},
     {MIDR_PART_T8140_TAHITI_PCORE, "A18 Pro Tahiti (P core)", NULL, &features_m4},
+    {MIDR_PART_T6050_SOTRA_MCORE, "M5 Pro Sotra (M core)", NULL, &features_m4},
+    {MIDR_PART_T6050_SOTRA_PCORE, "M5 Pro Sotra (P core)", NULL, &features_m4},
+    {MIDR_PART_T6051_SOTRAC_MCORE, "M5 Max Sotra C (M core)", NULL, &features_m4},
+    {MIDR_PART_T6051_SOTRAC_PCORE, "M5 Max Sotra C (P core)", NULL, &features_m4},
```
jannau, line 163: a little surprising that those have different MIDR part numbers, I was expected that M5 Pro and Max use the same CPU die. I guess this is one time programmable once they bond with the GPU die.

src/utils.h
```diff
@@ -335,7 +335,8 @@ static inline bool has_ecores(void)
 
 static inline int is_ecore(void)
 {
-    return has_ecores() && !(mrs(MPIDR_EL1) & (1 << 16));
+    u32 mpidr_el1 = mrs(MPIDR_EL1);
+    return has_ecores() && !(mpidr_el1 & (1 << 16)) && !(mpidr_el1 & (4 << 16));
```
jannau: shouldn't `has_ecores()` already return false for M5 Pro / Max? Do we need this at all? From a cursory look we're using `is_ecore()` only in places which we don't reach for apple sysreg locked devices.

yuyuyureka: Historically, I guess it went like this:

In old (pre-M1) SoCs which didn't have ecores **yet**, the nibble MPIDR_EL1[20:16] had been 0.
When they introduced ecores, the nibble was 0 for E-Cores and 1 for P-Cores. This means a value of 0 can mean we are on an old SoC (`!has_ecores()`) or we are on a different kind of core than E-Core.
So `is_ecore` is more like "isn't a P-Core and is new enough to have E-Cores", and I think we should add "is also not an M-Core" to that (as I've done here) _or_ we could check that the entire nibble is 0 (but I don't have enough data to support that this will work for all SoCs).
Either way, I don't think we should have to add all future models to the `has_ecores` list. It shouldn't be necessary, since we can distinguish by the respective MPIDR nibble here.

jannau: another option would be to rename `has_ecores()` to `is_heterogeneous()`

Please use `#define MIDR_CORE_TYPE_P BIT(16)` and `#define MIDR_CORE_TYPE_M BIT(18)` or at least `BIT(16)` and `BIT(18)` if we don't wnat to commit to names. I wonder what bit 17 is.

yuyuyureka: Done

src/smp.c
```diff
@@ -294,6 +294,8 @@ void smp_start_secondaries(void)
         case T6031:
         case T6034:
         case T6040:
+        case T6050:
+        case T6051:
```
svenpeter42, line 298: so i guess the offset is correct but there's some bug in iboot or whatever that prevents them from coming up?

src/utils.h
```diff
@@ -327,15 +327,19 @@ static inline void write64_lo_hi(u64 addr, u64 val)
...
 }
 
+#define MIDR_CORE_TYPE_P BIT(16)
+#define MIDR_CORE_TYPE_M BIT(18)
+
 static inline int is_ecore(void)
 {
-    return has_ecores() && !(mrs(MPIDR_EL1) & (1 << 16));
+    u32 mpidr_el1 = mrs(MPIDR_EL1);
+    return is_heterogeneous() && !(mpidr_el1 & MIDR_CORE_TYPE_P) && !(mpidr_el1 & MIDR_CORE_TYPE_M);
```
svenpeter42: i always find these longer && chains hard to read, maybe just three if statements instead?

src/utils.h
```diff
@@ -327,15 +327,19 @@ static inline void write64_lo_hi(u64 addr, u64 val)
...
 
 extern u32 board_id, chip_id;
-static inline bool has_ecores(void)
+static inline bool is_heterogeneous(void)
 {
     return !(chip_id == S5L8960X || chip_id == T7000 || chip_id == T7001 || chip_id == S8000 ||
              chip_id == S8001 || chip_id == S8003);
 }
 
+#define MIDR_CORE_TYPE_P BIT(16)
```
svenpeter42: these should go to midr.h

svenpeter42 (changes requested):

src/utils.h
yuyuyureka: Done

src/utils.h
yuyuyureka: Done

src/smp.c
yuyuyureka: Exactly

svenpeter42 (approved):

PR #614, opened 2026-07-03
adt: fix named_prop failure Error type

yuyuyureka:
If no property of this name exists, it should fail with `AdtError::NotFound`, but it attempted to run `next_property` on the last entry and thus returned `AdtError::BadOffset`.

### DIFF
--- a/rust/src/adt.rs
+++ b/rust/src/adt.rs
@@ -69,6 +69,76 @@ impl<'a> Iterator for ADTPropertyStringIterator<'a> {
     }
 }
 
+/// ADTPropertiesIterator
+#[derive(Debug)]
+pub struct ADTPropertiesIterator {
+    next_property_res: Option<Result<&'static ADTProperty, AdtError>>,
+    remaining: usize,
+}
+
+impl Iterator for ADTPropertiesIterator {
+    type Item = Result<&'static ADTProperty, AdtError>;
+
+    fn next(&mut self) -> Option<Result<&'static ADTProperty, AdtError>> {
+        let Some(property_res) = self.next_property_res.take() else {
+            return None;
+        };
+        let remaining = core::mem::take(&mut self.remaining);
+        let property = match property_res {
+            Err(e) => return Some(Err(e)),
+            Ok(v) => v,
+        };
+
+        if remaining > 1 {
+            self.next_property_res = Some(property.next_property());
+            self.remaining = remaining - 1;
+        }
+
+        Some(Ok(property))
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.remaining, Some(self.remaining))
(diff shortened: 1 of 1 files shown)


rust/src/adt.rs
```diff
@@ -400,13 +400,16 @@ impl ADTNode {
     /// Searches the node for a property with the given name, and returns it if
     /// found.
     pub fn named_prop(&self, name: &str) -> Result<&'static ADTProperty, AdtError> {
-        let mut p = self.first_property()?;
+        let mut p = self.first_property().unwrap();
```
jannau: `first_property()` can fail on not well formed (using this code as specification) ADT. Panics in m1n1's rust code are annoying so I'd prefer if we keep ``?` but at least it should be consistent with `self.first_property_mut()` in `named_prop_mut()`

rust/src/adt.rs
```diff
@@ -400,13 +400,16 @@ impl ADTNode {
...
+        if p.name() == name {
+            return Ok(p);
+        }
 
-        for _ in 0..self.property_count {
+        for _ in 0..(self.property_count - 1) {
```
jannau: `for _ in 1..self.property_count {` looks nicer and feels more natural

rust/src/adt.rs
```diff
@@ -400,13 +400,16 @@ impl ADTNode {
...
+            p = p.next_property().unwrap();
```
jannau: any reason for changing `p.next_property()?` to `p.next_property().unwrap()`

rust/src/adt.rs
```diff
@@ -400,13 +400,16 @@ impl ADTNode {
...
             if p.name() == name {
                 return Ok(p);
             }
-            p = p.next_property()?;
```
jannau, line 409: an option for keeping the existing structure and avoid duplicating the name check would be:
```
for i in 0..self.property_count {
    ...
    if i >= self.property_count - 1 {
        p = p.next_property()?;
    }
}
```
I would expect that this results in the same generated code as this change. I don't have a preference either they. A nicer alternative might be to provide an Iterator over node properties.

rust/src/adt.rs
yuyuyureka: Oops, that was a leftover from debugging. Thanks for catching it.

rust/src/adt.rs
yuyuyureka: Same here, debugging leftover

rust/src/adt.rs
yuyuyureka: I will rewrite it using iterators, as I don't like either of the variants.

yuyuyureka: Done

jannau (approved):

PR #616, opened 2026-07-03
proxyclient: skip reset_panic_counter on spmi gen 4

yuyuyureka:
I assume spmi gen 4 will not work for a while. This allows using the proxyclient with the devices which have spmi gen 4, with the drawback of not resetting the panic counter.


yuyuyureka:
Apparently adts from older macOS versions encode this differently:

14.8:
```
            compatible = [spmi,gen3]

```

26.5: 
```
            gen = 3 
            compatible = [aapl,spmi] 
```

yuyuyureka:
Closing in favor of #626

PR #618, opened 2026-07-07
PSCI via EFI

svenpeter42:


dkwo:
Does this feature require chainloading `u-boot` to work?

svenpeter42:
This probably doesn't do what you think it does. It will not result in better battery performance yet. It's just a possible way to upstream something that already exists in our downstream tree.

Right now it won't work at all with uboot.

dkwo:
Sorry if I misunderstood it. I was wondering whether (in the future) the improvements that this PR introduces will require the u-boot to work, but I think this is not the case. (Currently, chainloading kernel+initramfs as mini stage2 payload work well for me, and I was curious if that would be broken in the future, but it seems not.)

svenpeter42:
There are no improvements yet and it's probably even gonna be slightly worse right now. this only works without uboot and requires additional kernel patches. Eventually if this accepted upstream we'll make it work with uboot as well

PR #623, opened 2026-07-11
chickens: Move to their own subdirectory

svenpeter42:
They do not spark joy and we'll probably get a bunch more since new pre-M1 iDevices can boot unsigned coode now.


src/chickens/avalanche.c
```diff
@@ -1,7 +1,7 @@
 /* SPDX-License-Identifier: MIT */
 
-#include "cpu_regs.h"
-#include "utils.h"
+#include "../cpu_regs.h"
```
jannau: we could add `$srcdir/src` as `-I` dir to avoid using `../`. fine either way

jannau (approved):

src/chickens/avalanche.c
svenpeter42: good point

PR #624, opened 2026-07-14
hv: write CNTHCTL_EL2 after counter redirection

georgc4:
Write `CNTHCTL_EL2` only after the secondary CPU's  counter-redirection is configured. This only happens on M3+ because only these SOCs do counter redirection.

### DIFF
--- a/src/hv.c
+++ b/src/hv.c
@@ -216,7 +216,6 @@ static void hv_init_secondary(struct hv_secondary_info_t *info)
         msr(SYS_ACTLR_EL12, info->actlr_el1);
     else
         msr(SYS_IMP_APL_ACTLR_EL12, info->actlr_el1);
-    msr(CNTHCTL_EL2, info->cnthctl);
     if (cpu_features->apple_sysregs_unlocked) {
         msr(SYS_IMP_APL_SPRR_CONFIG_EL1, info->sprr_config);
         msr(SYS_IMP_APL_GXF_CONFIG_EL1, info->gxf_config);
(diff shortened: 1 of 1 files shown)


svenpeter42 (changes requested):
The fix seems correct but the reason needs to be adjusted: the CNTHCTL write has to be done after the counter redirection is configured and visible. I'd also add a comment before the isb to explain why it's there. That also explains why this only happens on M3+

svenpeter42:
please rebase to a single commit and drop the weird original explanation altogether

georgc4:
Squashed commits and retitled PR

svenpeter42:
LGTM now, just needs a rebase on the current main

georgc4:
Rebased

src/hv.c
```diff
@@ -229,6 +228,10 @@ static void hv_init_secondary(struct hv_secondary_info_t *info)
     if (cpu_features->apple_sysregs_unlocked)
         reg_mask(SYS_IMP_APL_CYC_OVRD, CYC_OVRD_WFI_MODE_MASK, CYC_OVRD_WFI_MODE(0));
 
+    //For M3 and up, CNTHCTL_EL2 must be written after the counter redirection
```
jannau: I suspect the format check complains over the missing space after `//`. `make format` should fix that if you have clang-format-20 installed. other versions might disagree over formating at a few other places

georgc4:
Added the space after the comment

PR #626, opened 2026-07-15
Support SPMI gen 4

yuyuyureka:
This allows reset_panic_counter (on the proxyclient side) and tps65989x (on the m1n1 side, in combination with #594) to work on M5 machines. Tested on T6050.

### DIFF
--- a/proxyclient/m1n1/hw/spmi.py
+++ b/proxyclient/m1n1/hw/spmi.py
@@ -1,7 +1,8 @@
 # SPDX-License-Identifier: MIT
 import struct
 
-from ..utils import *
+from .spmi1 import SPMI1Regs, R_CMD, R_REPLY
+from .spmi4 import SPMI4Regs
 
 __all__ = ["SPMI"]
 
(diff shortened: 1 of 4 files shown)


src/spmi.c
```diff
@@ -58,10 +103,22 @@ spmi_dev_t *spmi_init(const char *adt_node)
         return NULL;
     }
 
+    int gen;
+    const struct adt_property *gen_prop = adt_get_property(adt, adt_offset, "gen");
+    if (gen_prop)
+        gen = *(u64 *)gen_prop->value;
```
svenpeter42: i'd just use ADT_GETPROP here like we do almost everywhere else instead of this raw access

proxyclient/m1n1/hw/spmi.py
```diff
@@ -20,122 +21,20 @@
...
-
 class SPMI:
     def __init__(self, u, adt_path):
         self.u = u
         self.p = u.proxy
         self.iface = u.iface
         self.adt = u.adt[adt_path]
         self.base = self.adt.get_reg(0)[0]
-        self.regs = SPMIRegs(u, self.base)
+        gen = getattr(self.adt, "gen") if hasattr(self.adt, "gen") else -1
```
svenpeter42: nit: could just be `gen = getattr(self.adt, "gen", -1)`

svenpeter42 (changes requested):

src/spmi.c
yuyuyureka: Done and, since adt_getprop_copy exposes the error code, added distinguishing between NotFound (treat like version <= 3) and other errors (bubble up the error).

proxyclient/m1n1/hw/spmi.py
yuyuyureka: Done

src/spmi.c
```diff
@@ -58,10 +103,24 @@ spmi_dev_t *spmi_init(const char *adt_node)
         return NULL;
     }
 
+    u32 gen;
+    int ret = ADT_GETPROP(adt, adt_offset, "gen", &gen);
+    if (ret == -1) // NotFound
+        gen = -1;
```
svenpeter42, line 109: gen is unsigned, so when this isn't found we'll get gen = 0xffffffff >= 4 and then dev->regs = &regs_gen4, won't we? so we either need gen = 0 here or just use s32 for gen.

yuyuyureka: Good catch! changed type of `gen` to `s32`

PR #627, opened 2026-07-15
Some apfs experiments

WhatAmISupposedToPutHere:


svenpeter42 (approved):
python looks reasonable to me, the rust side is probably fine for experiments as well but if we ever need gigalocker or APFS support inside m1n1 we probably want to turn all those assert, etc. panics into just returning an error instead

jannau (approved):

PR #630, opened 2026-07-18
DCP: Add initial 14.7/14.8.3 ABI

chadmed:
Still some stuff to figure out, but with this information Linux can get all the way up to trying to submit swaps.


proxyclient/hv/trace_dcp.py
```diff
@@ -810,22 +812,8 @@ def call(self, msg, dir):
...
-            print(f"Message: {tag} ({KNOWN_MSGS.get(tag, 'unk')}): (in {in_len:#x}, out {out_len:#x})")
-            if data_in:
-                print(f"{dir} Input ({len(data_in):#x} bytes):")
-                chexdump(data_in[:self.state.max_len])
-
-        #if tag not in KNOWN_MSGS:
-            #hv.run_shell()
-
-        if self.state.dumpfile:
-            dump = f"CALL {dir} {msg.value:#018x} {self.name} {state.off:#x} {state.tag} {in_len:#x} {out_len:#x} {data_in.hex()}\n"
```
svenpeter42, line 826: maybe I just missed where you moved this, but do you still write the CALLs to the dumpfile?

chadmed: The IPC client logs calls itself, and with 8f4eb15631b2 we can simply pass the tracer's `self.log` in to the IPC client to dump out to the HV log file if `-l` was specified when running the guest. This keeps everything in one place (excluding ACKs which I will get around to making work the same as CALLs at some point)

PR #633, opened 2026-07-21
pcie: Add t8132/t6040 support

yuyuyureka:
APCIE_PHY_CTRL_RESET moves from BIT(7) to BIT(4). Otherwise the same as T8122.

### DIFF
--- a/src/pcie.c
+++ b/src/pcie.c
@@ -35,12 +35,13 @@
 
 /* PHY registers */
 
-#define APCIE_PHY_CTRL         0x000
-#define APCIE_PHY_CTRL_CLK0REQ BIT(0)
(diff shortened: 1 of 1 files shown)


src/pcie.c
```diff
@@ -449,7 +472,10 @@ static int pcie_init_controller(int controller, const char *path)
             return -1;
         }
 
-        clear32(state->phy_base[phy] + APCIE_PHY_CTRL, APCIE_PHY_CTRL_RESET);
+        if (state->pcie_regs->type == APCIE_T8132)
```
svenpeter42: I'd just add the bit to struct reg_info

svenpeter42 (changes requested):

svenpeter42 (approved):

PR #635, opened 2026-07-22
memory: Introduce proper 4-level paging

asdfugil:
Introduce proper 4-level paging to reduce the size of the pagetables on 4K devices, which used to statically allocate a 4 MiB pagetable.

Should also slightly reduce memory consumption on 16K size the upper half of the VA space is never used in practice.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -60,64 +60,41 @@ static inline void write_sctlr(u64 val)
 
 #define VADDR_L3_INDEX_BITS (is_16k() ? 11 : 9)
 #define VADDR_L2_INDEX_BITS (is_16k() ? 11 : 9)
-// We treat two concatenated L1 page tables as one on 16K
-// And all the L1 page tables together on 4K...
-#define VADDR_L1_INDEX_BITS        (is_16k() ? 12 : 18)
-#define VADDR_L1_INDEX_BITS_ACTUAL (is_16k() ? 11 : 9)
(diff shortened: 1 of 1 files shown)


src/memory.c
```diff
@@ -288,21 +311,40 @@ int mmu_map(u64 from, u64 to, u64 size)
...
+    if (chunk) {
+        mmu_pt_map_l2(from, to, chunk);
+        from += chunk;
+        to += chunk;
+        size -= chunk;
+    }
+
+    // L1 mappings
+    chunk = ALIGN_DOWN(size, MASK(VADDR_L1_OFFSET_BITS));
+    if (chunk && (to & VADDR_L1_ALIGN_MASK) == 0) {
```
svenpeter42: I think this needs a !is_16k guard because adding blocks like this isn't valid with 16K pages I think. not sure if this can actually be reached in practice though

svenpeter42 (changes requested):

src/memory.c
asdfugil: Right, good catch

https://developer.arm.com/documentation/ddi0487/mc/-Part-D-The-AArch64-System-Level-Architecture/-Chapter-D8-The-AArch64-Virtual-Memory-System-Architecture/-D8-2-Translation-process/-D8-2-9-VMSAv8-64-translation-using-the-16KB-granule

PR #636, opened 2026-07-23
tps6598x,usb: factor out hpm iteration

yuyuyureka:
Tested on T6000


yuyuyureka:
Addressed a comment raised by chaos_princess on IRC: added #defines for the return values

PR #637, opened 2026-07-24
nvme: fix cdw12 value to read 1 block

yuyuyureka:
This is required for newer firmwares (26.x) at least on both M3 and M4.


asdfugil (approved):
This is also required on A11.

rhelvista:
with that also M5

PR #639, opened 2026-07-25
kboot_atc: t8132 support (like t8122)

yuyuyureka:


src/kboot_atc.c
```diff
@@ -278,6 +278,7 @@ static const struct atc_fuse_info atc_fuses_t8112_port1[] = {
 // strings in ADT atc-phy nodes.
 static const struct atc_fuse_hw atc_fuses[] = {
     {"atc-phy,t8122", -1, NULL, 0},
+    {"atc-phy,t8132", -1, NULL, 0},
```
jannau, line 280: please read the comment above. this line should be above `"atc-phy,t8122"`, not that it should matter for SOCs without fuses

jannau (changes requested):

PR #641, opened 2026-07-26
WIP: kboot_atc: t8130 support (like t8122, but no usb4)

yuyuyureka:
The tunable names match t8122, but USB 3 does not actually work with the t8122 atcphy driver, so this is WIP

### DIFF
--- a/src/kboot_atc.c
+++ b/src/kboot_atc.c
@@ -146,6 +146,39 @@ static const struct adt_tunable_info atc_tunables_t8122[] = {
     {"tunable_LN1_TX_SHM_CIO_DFLT", "apple,tunable-lane1-cio", 0x14000, 0x1000, true},
 };
 
+static const struct adt_tunable_info atc_tunables_t8130[] = {
+    {"tunable_ATC0AXI2AF", "apple,tunable-axi2af", 0x0, 0x8000, true},
+    {"tunable_ATC_FABRIC", "apple,tunable-common-b", 0x44000, 0x4000, true},
+
+    {"tunable_CIO3PLL_CORE", "apple,tunable-common-b", 0x2a00, 0x200, true},
+    {"tunable_CIO3PLL_TOP", "apple,tunable-common-b", 0x2800, 0x200, true},
+    {"tunable_ACIOPHY_LANE_USBC0", "apple,tunable-common-b", 0x5000, 0x1000, true},
+    {"tunable_ACIOPHY_PLL_TOP", "apple,tunable-common-b", 0x1000, 0x4000, true},
+    {"tunable_ACIOPHY_TOP", "apple,tunable-common-b", 0x0, 0x4000, true},
+
+    {"tunable_AUSCMN_DIG", "apple,tunable-common-b", 0x800, 0x200, true},
+    {"tunable_AUSPLL_CORE", "apple,tunable-common-b", 0x2200, 0x4000, true},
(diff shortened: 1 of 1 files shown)


src/kboot_atc.c
```diff
@@ -123,23 +123,23 @@ static const struct adt_tunable_info atc_tunables_t8122[] = {
     {"tunable_LN0_RX_SHM_USB_DFLT", "apple,tunable-lane0-usb", 0xb000, 0x1000, true},
     {"tunable_LN0_TX_TOP_USB_DFLT", "apple,tunable-lane0-usb", 0xc000, 0x1000, true},
     {"tunable_LN0_TX_SHM_USB_DFLT", "apple,tunable-lane0-usb", 0xd000, 0x1000, true},
-    {"tunable_LN0_RX_TOP_CIO_DFLT", "apple,tunable-lane0-cio", 0x9000, 0x1000, true},
-    {"tunable_LN0_RX_EQ_CIO_EQA", "apple,tunable-lane0-cio", 0xa000, 0x1000, true},
-    {"tunable_LN0_RX_SHM_CIO_DFLT", "apple,tunable-lane0-cio", 0xb000, 0x1000, true},
-    {"tunable_LN0_TX_TOP_CIO_DFLT", "apple,tunable-lane0-cio", 0xc000, 0x1000, true},
-    {"tunable_LN0_TX_SHM_CIO_DFLT", "apple,tunable-lane0-cio", 0xd000, 0x1000, true},
+    {"tunable_LN0_RX_TOP_CIO_DFLT", "apple,tunable-lane0-cio", 0x9000, 0x1000, false},
```
jannau: I don't making tunables optional because they are no longer present on a new SoC is a good idea. Please add `atc_tunables_t8130` and just remove the lines

src/kboot_atc.c
```diff
@@ -369,6 +371,11 @@ static int dt_append_atc_tunable(void *dt, int adt_node, int fdt_node,
     const struct atc_tunable *tunable_adt =
         adt_getprop(adt, adt_node, tunable_info->adt_name, &tunables_len);
 
+    // Alias tunable_ATC0AXI2AF to tunable_ATCAXI2AF, seen on M4 Mac mini (J773g)
+    if (!tunable_adt && !strcmp(tunable_info->adt_name, "tunable_ATC0AXI2AF")) {
```
jannau, line 412: is this just on t8132 or on t8130 as well? if it is on both this should become unnecessary with its own tunable struct

jannau (changes requested):

src/kboot_atc.c
yuyuyureka: No, only t8132. And this part is already merged, I just have to rebase.

src/kboot_atc.c
```diff
@@ -453,8 +487,11 @@ static void dt_copy_atc_tunables(void *dt, const char *adt_path, const char *dt_
         goto cleanup;
     }
 
-    if (adt_is_compatible_at(adt, adt_node, "atc-phy,t8122", 0) ||
-        adt_is_compatible_at(adt, adt_node, "atc-phy,t8132", 0)) {
+    if (adt_is_compatible_at(adt, adt_node, "atc-phy,t8130", 0)) {
```
jannau, line 490: diff is easier to read with `} else if (adt_is_compatible_at(adt, adt_node, "atc-phy,t8130", 0)) {`

PR #643, opened 2026-07-28
SPRR/GXF emulation for the HV

svenpeter42:


src/hv.c
```diff
@@ -218,7 +218,8 @@ static void hv_init_secondary(struct hv_secondary_info_t *info)
         msr(SYS_IMP_APL_ACTLR_EL12, info->actlr_el1);
     if (cpu_features->apple_sysregs_unlocked) {
         msr(SYS_IMP_APL_SPRR_CONFIG_EL1, info->sprr_config);
-        msr(SYS_IMP_APL_GXF_CONFIG_EL1, info->gxf_config);
+        if (info->sprr_config & SPRR_CONFIG_EN)
```
jannau, line 222: this could use the commit message as comment

jannau (commented):
single nit for for the preparation changes. Feel free to open a separate PR for those so we can get them out of the way.
I glanced so far only at the SPRR/GXF emulation itself

PR #644, opened 2026-08-01
rust: apfs: Build apfs only for feature chainload

jannau:
Quick build fix as the rust GPT/NVMe code depends on fatfs. This needs to be sorted out properly when gigalocker is used for more than experiments. Easiest solution would probably to just drop the chainload cargo feature but retain the CHAINLOADING make variable / define to disable chainloading in m1n1.

Fixes: 07c2c3af ("proxyclient: Add get_gigalocker")

### DIFF
--- a/rust/src/lib.rs
+++ b/rust/src/lib.rs
@@ -12,6 +12,7 @@
 extern crate alloc;
 
 pub mod adt;
+#[cfg(feature = "chainload")]
 pub mod apfs;
 #[cfg(feature = "chainload")]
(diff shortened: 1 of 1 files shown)


WhatAmISupposedToPutHere:
Unlikely to be used for more than experiments, as doing sep handoff from m1n1 all the way to linux will be rather annoying. So, hanging it off CHAINLOADING is more than fine

WhatAmISupposedToPutHere (approved):

PR #648, opened 2026-08-02
nvme: skip LINEAR_SQ_CTRL & UNKNOWN_CTRL writes for FW >= 15.0

yuyuyureka:
These setup steps are not necessary on recent firmware. Tested `nvme_init()` and `nvme_read()` on M3 Max with 15.7.8 & 26.5.1 FW.


svenpeter42:
this breaks M1 on firmware 27 at least. i'll try to figure out what that one actually needs

PR #651, opened 2026-08-06
NVMe fixes for macOS 15+

svenpeter42:

PR #652, opened 2026-08-06
nvme: M4+ support

yuyuyureka:
Depends on #651

### DIFF
--- a/src/nvme.c
+++ b/src/nvme.c
@@ -42,6 +42,8 @@
 #define NVME_LINEAR_SQ_CTRL    0x24908
 #define NVME_LINEAR_SQ_CTRL_EN BIT(0)
 
+#define NVME_IOQ_CMDS           0x1200
+#define NVME_IOQ_CQES           0x1208
 #define NVME_MAX_PEND_CMDS_CTRL 0x1210
 #define NVME_DB_LINEAR_ASQ      0x2490c
 #define NVME_DB_LINEAR_IOSQ     0x24910
@@ -120,6 +122,11 @@ static_assert(sizeof(struct nvme_command) == 64, "invalid nvme_command size");
 static_assert(sizeof(struct nvme_completion) == 16, "invalid nvme_completion size");
 static_assert(sizeof(struct apple_nvmmu_tcb) == 128, "invalid apple_nvmmu_tcb size");
 
+static enum {
+    NVME_T8103 = 0,
+    NVME_T8132 = 1,
+} nvme_type;
+
 static bool nvme_initialized = false;
 static u8 nvme_die;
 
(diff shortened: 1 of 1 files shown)


src/nvme.c
```diff
@@ -307,10 +310,24 @@ bool nvme_init(void)
...
         return NULL;
     }
+
+    if (adt_get_reg(adt, adt_path, "reg", 3, &nvmmu_base, NULL) < 0) {
+        printf("nvme: Error getting NVMMU base address.\n");
+        return NULL;
+    }
+    if (reg_len >= 10) {
+        if (adt_get_reg(adt, adt_path, "reg", 9, &nvme_base, NULL) < 0) {
+            printf("nvme: Error getting NVMMU base address.\n");
```
jannau: "NVMe base address"

src/nvme.c
```diff
@@ -307,10 +310,24 @@ bool nvme_init(void)
...
+            return NULL;
+        }
+    } else
+        nvme_base = nvmmu_base;
```
jannau: please use braces for all or none branches. I wonder if we can add this to our clang-format config

src/nvme.c
```diff
@@ -307,10 +310,24 @@ bool nvme_init(void)
...
+    if (!adt_getprop(adt, node, "reg", &reg_len)) {
+        printf("nvme: Error getting reg length.\n");
         return NULL;
     }
+
+    if (adt_get_reg(adt, adt_path, "reg", 3, &nvmmu_base, NULL) < 0) {
+        printf("nvme: Error getting NVMMU base address.\n");
+        return NULL;
+    }
+    if (reg_len >= 10) {
```
jannau: this needs at least a comment but I think it would be clearer if we define our own `ans`-generation enum and assign it based on chip_id unless there is something more suitable in the ADT. The ans nodes unfortunately don't have anything which looks useful.
Let's hope this isn't something Apple (can) change via firmware update.

yuyuyureka: There seems to be [one property](https://oftc.catirclogs.org/asahi-re/2026-07-02#35467572) which is present on M4+ devices. I'll try to use that instead of the reg count, since the reg count turns out to be >= 10 on M2 devices too: https://termbin.com/32tz

svenpeter42: I'd honestly also just key it off the chip id at this point. nvme-secure-bar is the same situation: needs a comment at least.

yuyuyureka: For $reasons, I'm pretty sure nvme-secure-bar is the correct thing to check for here. Added more comments to explain the differences between the generations.

svenpeter42 (approved):

src/nvme.c
svenpeter42: $reasons?

PR #653, opened 2026-08-07
mcc: Support SLC and TZ carveouts on T8122 and T6030 

jannau:
```
mcc: Add support for T8122 SLC and carveouts

T8122's mcc appears to have a single cache enable register instead of
per plane ones. Add support for this and fill all other information.


mcc: Add support for T6030 SLC and carveouts

T6030's mcc is identical to T8122 with 50% more planes and dcs. Only
it's TZ carveout offsets differ. Since they differ as well from T6031
rename t603x_tz_regs to t6031_tz_regs.
```

### DIFF
--- a/src/mcc.c
+++ b/src/mcc.c
@@ -47,6 +47,14 @@ struct tz_regs t8103_tz_regs = {
     .enable = 0x6a8,
 };
 
+struct tz_regs t8122_tz_regs = {
+    .count = 4,
+    .stride = 0x14,
+    .start = 0x6cc,
+    .end = 0x6d0,
+    .enable = 0x6d8,
+};
+
 struct tz_regs t602x_tz_regs = {
     .count = 4,
     .stride = 0x14,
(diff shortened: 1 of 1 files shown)


src/mcc.c
```diff
@@ -66,6 +82,9 @@ struct tz_regs t603x_tz_regs = {
 #define PLANE_CACHE_ENABLE 0x1c00
 #define PLANE_CACHE_STATUS 0x1c04
 
+// T8122 does not seem to have a per plane enable but enable caches via single write to 0x1c1c00
+#define T8122_CACHE_ENABLE 0x1c1c00
```
WhatAmISupposedToPutHere: This appears to be unused

src/mcc.c
```diff
@@ -111,6 +136,7 @@ struct mcc_carveout mcc_carveouts[PLANE_TZ_MAX_REGS + 1];
 struct mcc_regs {
     u64 plane_base;
     u64 plane_stride;
+    u64 global_cache_enable;
```
WhatAmISupposedToPutHere: Looks like this never gets written?

jannau: supposed to be set to `T8122_CACHE_ENABLE`. It appears that the per plane enable writes still work. Consistent with SLC working on the M3 Max. This is just not what macOS 14.8.3 does on T8122 / T6030 and probably T6031.
This got lost in refactoring and I did not notice because the SLC was enabled. I'll remove it and the changes in `mcc_enable_cache()` and amend the commit message instead.

WhatAmISupposedToPutHere (approved):

PR #657, opened 2026-08-14
kboot: add reserved M4 smp-ro region

yuyuyureka:
From experimentation, s3_0_c11_c0_0 contains the first and s3_0_c11_c0_1 the last 4K page which is read-only to secondary cores, and these are 0xb000 / 44KiB apart on M4 / M4 Pro / A18 Pro.

It is unclear if this region serves a special purpose, but to avoid writing to it accidentally, add a reserved-memory node of 48KB to the fdt and assert that s3_0_c11_c0_1 is contained in it.

In combination with disabling WFI/WFIT instructions, this makes smp boot reliable on M4 / M4 Pro.

Discovered-by: Niklas Sheth <hi@niklassheth.com>

I take suggestions for names for these sysregs. It is difficult to name them given we do not know their purpose.


svenpeter42:
is this only read-only for the *secondary* cores? because that sounds rather odd.

might be CTRR, are you sure there's no memory region that matches this in the ADT?

yuyuyureka:
> is this only read-only for the _secondary_ cores? because that sounds rather odd.

The boot core can write to this region. It is read-only only for the secondary cores, as in cores other than the boot core. 

> might be CTRR, are you sure there's no memory region that matches this in the ADT?

I could not find any matches for both hex and decimal representations in my adt.

svenpeter42:
sounds like another instance of these regs here: https://github.com/AsahiLinux/m1n1/blob/5f00537804080932450142e80cabedb9a9bf9ead/proxyclient/m1n1/apple_regs.json#L136 
so maybe just call them CTRR_B_... or CTRR_M4_.. or CTRR_T8122_... or similar

svenpeter42:
wait, it's read-only for the secondary cores but the sysreg is also set on the primary? that sounds like something else than CTRR then

yuyuyureka:
Dumped (3, 0, 11, *, *) from all cores: https://termbin.com/0arok

Diffing core 6 (boot core) and core 7:
foo


As you can see, s3_0_c11_c1_4 and s3_0_c11_c2_0 differ between the cores that can write to the area and those which can not, and are probably some kinds of enable regs. The effectively read-only range in this run is 0x10003e44000 - 0x10003e4ffff, and this range (and the reg values) change a bit on every boot.

yuyuyureka:
Changed to use s3_0_c11_c1_*, since this is where the enable reg which differs between boot and non-boot cores  lives.
Gave the regs the names `CTRR_M4_{LWR,UPR,CTL}`, and added them to `apple_regs.json` and `cpu_regs.h`.

src/kboot.c
```diff
@@ -2801,6 +2771,18 @@ int kboot_prepare_dt(void *fdt)
     if (fdt_add_mem_rsv(dt, (u64)_base, ((u64)_end) - ((u64)_base)))
         bail("FDT: couldn't add reservation for m1n1\n");
 
+    if (!cpu_features->apple_sysregs_unlocked) {
+        // On M4* / A18 Pro / M5*, there is a CTRR instance left enabled only
+        // on the secondary (non-boot) cores. Add a reserved-memory region to
+        // prevent Linux from crashing when allocating this memory and writing
+        // to it from a secondary core.
+        u64 ro_base = mrs(CTRR_M4_LWR_EL2);
+        u64 ro_size = 12 * SZ_4K;
```
svenpeter42: shouldn't this just be the upper reg minus the lower one instead of hardcoding it here?

yuyuyureka: We can do upper - lower + SZ_4K (since upper reg contains the beginning of the last RO page).
My worry was that if this changes (the covered size gets dramatically bigger or something) we should double check if this is still left enabled on the secondary cores before we adjust the size of the reservation. But if you say we should just reserve it either way, I can also do that.

svenpeter42: The assert below is gonna crash even with the current code if upper is too large as well, isn't it? I'd just reserve ALIGN_UP(upper - lower, SZ_4K) or possibly even SZ_16K if that's the CTRR granularity and maybe add a warning if that size is > some value. This will only change with firmware updates anyway.

yuyuyureka: > The assert below is gonna crash even with the current code if upper is too large as well, isn't it?

The assert would have thrown if the size is too large, yeah. That was its intended purpose.

CTRR granularity is SZ_4K. Changed it to align the end up to SZ_4K, and then take the difference as size and print a warning if it's larger than 48KB (what we're currently seeing).

src/kboot.c
```diff
@@ -2801,6 +2771,19 @@ int kboot_prepare_dt(void *fdt)
...
         bail("FDT: couldn't add reservation for m1n1\n");
 
+    if (!cpu_features->apple_sysregs_unlocked) {
+        // On M4* / A18 Pro / M5*, there is a CTRR instance left enabled only
+        // on the secondary (non-boot) cores. Add a reserved-memory region to
+        // prevent Linux from crashing when allocating this memory and writing
+        // to it from a secondary core.
+        u64 ro_start = mrs(CTRR_M4_LWR_EL2);
+        u64 ro_end = ALIGN_UP(mrs(CTRR_M4_UPR_EL2), SZ_4K);
+        if (ro_start - ro_end > 12 * SZ_4K)
```
svenpeter42: this is the wrong way around, isn't it?

yuyuyureka: Yes, I should've taken the time to test it before I pushed :disappointed:

yuyuyureka:
the mrs fails in EL1, so this breaks m1n1-in-m1n1 hv

PR #661, opened 2026-08-15
sart: add SARTv4 support

yuyuyureka:
As seen on M5 Pro/Max (T6050)

When combined with #652, this makes nvme_init() and nvme_read() work on T6050


src/sart.c
```diff
@@ -151,6 +163,35 @@ static bool sart3_set_entry(sart_dev_t *sart, int index, u8 flags, void *paddr_,
     return true;
 }
 
+static void sart4_get_entry(sart_dev_t *sart, int index, u8 *flags, void **paddr, size_t *size)
```
WhatAmISupposedToPutHere, line 166: that looks exactly like sart3_ except for the offsets, right? Maybe merge them somehow?

yuyuyureka: Just like sart2 is a copy of sart0 with different masks/offsets.

WhatAmISupposedToPutHere: Arguably a reason to merge those too :P. But up to you.

yuyuyureka: done at https://github.com/yuyuyureka/m1n1/commit/6af0f26f768a7b3d891079b9a5aa6df934474b94, but I would appreciate if that could be reviewed separately

WhatAmISupposedToPutHere (approved):

PR #662, opened 2026-08-15
sart: combine SARTv{0,2} and SARTv{3,4}

yuyuyureka:
Followup to #661

### DIFF
--- a/src/sart.c
+++ b/src/sart.c
@@ -12,67 +12,35 @@ struct sart_dev {
     /* This is probably a bitfield but the exact meaning of each bit is unknown. */
     u32 flags_allow;
 
+    u64 config_offset;
+    u64 config_stride;
+    u64 config_size_max;
+    u64 paddr_offset;
+    u64 paddr_stride;
+    u64 paddr_shift;
+    u64 size_shift;
+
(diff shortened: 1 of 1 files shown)


src/sart.c
```diff
@@ -12,182 +12,89 @@ struct sart_dev {
     /* This is probably a bitfield but the exact meaning of each bit is unknown. */
     u32 flags_allow;
```
svenpeter42, line 13: the flags are always 0xff so we may as well just drop that field here

svenpeter42 (approved):
I'm not against this, but I also don't think that it buys us all that much.

src/sart.c
yuyuyureka: They're `0xf` for SARTv0

yuyuyureka:
I also don't have a strong opinion on whether this is better than previously. @WhatAmISupposedToPutHere suggested it, so maybe you two discuss it ;)

jannau:
I don't have a strong opinion but I would prefer not to merge this.

It might be worthwhile to just add `{config,padr,size}_offset` to `sart_dev` and combine SARTv3 and SARTv4 that way.

PR #668, opened 2026-08-23
kboot: Use new ACIO aliases

svenpeter42:
The device tree checking tools don't like the aliases with '_'


jannau (approved):
There can be an argument made that the old style tunable should be tested first so that `dt_copy_acio_tunables()` complains that the upstream alias is missing. Doesn't matter much though.

PR #672, opened 2026-08-29
kboot: disable wfi/wfit if wfi loses state

yuyuyureka:


src/kboot.c
```diff
@@ -316,6 +316,11 @@ static int dt_set_chosen(void)
     if (dt_set_rng_seed_sep(node))
         return dt_set_rng_seed_adt(node);
 
+    if (wfi_loses_state()) {
+        if (fdt_appendprop_string(dt, node, "bootargs", " idle=nop arm64.nowfxt"))
```
jannau, line 320: does this actually work? in my test with " nohibernate" this resulted in
```
bootargs = "root=...", " nohibernate";
```
and the kernel ignored the second string. At least in its `Kernel command line: ` debug output. I think the issue is that the existing string still has a terminating `\0`.

src/kboot.c
```diff
@@ -316,6 +316,27 @@ static int dt_set_chosen(void)
     if (dt_set_rng_seed_sep(node))
         return dt_set_rng_seed_adt(node);
 
+    if (wfi_loses_state()) {
+        char *to_append = " idle=nop arm64.nowfxt";
```
jannau: `const char to_append[] =" idle=nop arm64.nowfxt";`

src/kboot.c
```diff
@@ -316,6 +316,27 @@ static int dt_set_chosen(void)
...
+
+        /*
+         * fdt_appendprop_string would create a nother property values:
```
jannau: stray space between "a" and "nother"

src/kboot.c
```diff
@@ -316,6 +316,27 @@ static int dt_set_chosen(void)
...
+         * "a", "b". But we want "ab", so manually create or grow the prop,
+         * then overwrite the \0 of the existing string.
+         */
+        int oldlen, newlen;
+        void *prop_data;
+
+        if (!fdt_get_property(dt, node, "bootargs", &oldlen) || oldlen < 1)
+            oldlen = 1;
+        newlen = oldlen + strlen(to_append);
```
jannau: `sizeof(to_append) - 1`

src/kboot.c
```diff
@@ -316,6 +316,27 @@ static int dt_set_chosen(void)
...
+
+        if (fdt_setprop_placeholder(dt, node, "bootargs", newlen, &prop_data))
+            bail("FDT: failed to append to 'bootargs' property\n");
+
+        memcpy(prop_data + oldlen - 1, to_append, strlen(to_append) + 1);
```
jannau: `strncpy(prop_data + oldlen - 1, to_append, sizeof(to_append))`

jannau (commented):
not sure which string manipulation variants I dislike more. I prefer arrays over pointers since it allows `sizeof` but I'm not sure if it helps much here.

Is there a reason why this is still a draft?

yuyuyureka:
> Is there a reason why this is still a draft?

Mainly because I haven't gotten around to testing it yet

